PR #16, opened 2020-04-15
Fix failing CI testing suite

AObuchow:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,7 @@ pipeline{
     stage("Maven Build"){
         steps {
           withMaven {
-            sh './mvnw -B verify --file lemminx-maven/pom.xml -DskipTests'
+            sh './mvnw -B verify --file lemminx-maven/pom.xml'
           }
         }
     }
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -1,4 +1,6 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>org.eclipse.lemminx</groupId>
@@ -86,6 +88,17 @@
 			<plugin>
 				<groupId>org.apache.felix</groupId>
 				<artifactId>maven-bundle-plugin</artifactId>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.22.2</version>
+				<configuration>
+					<systemPropertyVariables>
+						<maven.repo.local>${maven.repo.local}</maven.repo.local>
+						<org.eclipse.lemminx.maven.searcher.RemoteRepositoryIndexSearcher.disableCentralIndex>false</org.eclipse.lemminx.maven.searcher.RemoteRepositoryIndexSearcher.disableCentralIndex>
+					</systemPropertyVariables>
+				</configuration>
 			</plugin>
 		</plugins>
 	</build>
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
@@ -56,11 +56,14 @@ import org.codehaus.plexus.PlexusContainer;
 import org.eclipse.aether.repository.RemoteRepository;
 
 public class RemoteRepositoryIndexSearcher {
+	private static final String MAVEN_INDEX_DIRECTORY = "._maven_index_";
 	private static final String PACKAGING_TYPE_JAR = "jar";
 	private static final String PACKAGING_TYPE_MAVEN_PLUGIN = "maven-plugin";
 
 	public static final RemoteRepository CENTRAL_REPO = new RemoteRepository.Builder("central", "default", "https://repo.maven.apache.org/maven2").build();
 	private final Set<RemoteRepository> knownRepositories;
+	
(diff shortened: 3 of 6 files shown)


fbricon:
commit message and actual change are unrelated

AObuchow:
Yes this is true as the .m2 repo location change is on the Jenkins
configuration side. I’ll update the commit description to be more accurate.
In reality, this is just a test PR that will be discarded and not merged.

lemminx-maven/pom.xml
```diff
@@ -1,9 +1,11 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>org.eclipse.lemminx</groupId>
 	<artifactId>lemminx-maven</artifactId>
-	<version>0.0.1-SNAPSHOT</version>
+	<version>0.0.3-SNAPSHOT</version>
```
AObuchow, line 8: Change back to 0.0.1

Jenkinsfile
```diff
@@ -6,12 +6,13 @@ pipeline{
   environment {
     MAVEN_HOME = "$WORKSPACE/.m2/"
     MAVEN_USER_HOME = "$MAVEN_HOME"
+    MAVEN_REPOSITORY = "$MAVEN_HOME/repository/"
```
AObuchow, line 9: Generalize things so that we can just use MAVEN_USER_HOME instead of having to point to a /repository/ directory.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
```diff
@@ -62,6 +62,8 @@
 	public static final RemoteRepository CENTRAL_REPO = new RemoteRepository.Builder("central", "default", "https://repo.maven.apache.org/maven2").build();
 	private final Set<RemoteRepository> knownRepositories;
 	
+	boolean disableCentralIndex = System.getProperty(getClass().getName() + ".disableCentralIndex") != null;
```
AObuchow, line 65: @Mickaelistria this might be an ugly way to get a boolean from a system property, since it doesn't take into account the value passed (only null or not). Should I use a string and then convert it's value into a boolean?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
```diff
@@ -214,6 +218,12 @@ public void transferCompleted(TransferEvent transferEvent) {
 		if (context == null) {
 			return CompletableFuture.runAsync(() -> { throw new IllegalArgumentException("context mustn't be null"); });
 		}
+		if ((context.getId().equals("https://repo.maven.apache.org/maven2") || context.getId().equals("central"))
+				&& disableCentralIndex) {
+			return CompletableFuture.runAsync(() -> {
+				System.out.println("Central repository index disabled");
```
AObuchow, line 224: @mickaelistria do you have a suggestion for what the completable future should return instead of just printing something to stdout?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
```diff
@@ -214,6 +218,12 @@ public void transferCompleted(TransferEvent transferEvent) {
 		if (context == null) {
 			return CompletableFuture.runAsync(() -> { throw new IllegalArgumentException("context mustn't be null"); });
 		}
+		if ((context.getId().equals("https://repo.maven.apache.org/maven2") || context.getId().equals("central"))
```
AObuchow, line 221: We could also check eqality for the `CENTRAL_REPO` instead of doing `context.getId().equals("central")`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/searcher/RemoteRepositoryIndexSearcher.java
AObuchow: Ok will do.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/ClientServerConnection.java
```diff
@@ -72,7 +74,12 @@ public ClientServerConnection() throws IOException {
 	}
 
 	public void stop() throws InterruptedException, ExecutionException {
-		languageServer.shutdown().get();
+		try {
+			languageServer.shutdown().get(1, TimeUnit.MINUTES);
```
AObuchow, line 78: This was added just as a test and probably should be removed

AObuchow: Adding this forced shutdown seems to stop the CI from going on forever. It looks like the language server isin't able to shutdown correctly. (This is most likely due to the shutdown of the extension not finishing properly, not the language server's fault)

lemminx-maven/pom.xml
```diff
@@ -87,6 +89,17 @@
...
 				<artifactId>maven-bundle-plugin</artifactId>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.22.2</version>
+				<configuration>
+					<systemPropertyVariables>
+						<maven.repo.local>${maven.repo.local}</maven.repo.local>
+						<org.eclipse.lemminx.maven.searcher.RemoteRepositoryIndexSearcher.disableCentralIndex>false</org.eclipse.lemminx.maven.searcher.RemoteRepositoryIndexSearcher.disableCentralIndex>
```
mickaelistria, line 99: For the disableCentralIndex, as this is also desired for any testing (not only Maven), we could consider building a TestSuite referencing all tests and setting this as part of the test suite.

AObuchow:
> I'll split this into 2 different patches: one to disable central Index for testing (in all cases) so test should work easily on any terminal. And another where I'll try to run them from the Jenkins.

Sounds good with me.

lemminx-maven/pom.xml
AObuchow: I'm not sure I fully understand - I was under the impression that disableCentralIndex was already affecting the entire test suite? I guess it'll be easier to understand once you show me the new patch :)

PR #21, opened 2020-04-21
Add support for Maven Property Hover

### LINKED ISSUE #13: Add line break to Maven property hover
The hover for Maven properties requires a line break after the property name, so that it reads more clearly.

### LINKED ISSUE #20: Maven property hover missing?
For some reason, Maven property hover is missing. It was originally added in https://github.com/AObuchow/lsp4xml-extensions-maven/commit/8ab67276e89506e71cb2b78bede86f6294f507cc

AObuchow:
Fix #20
Fix #13 


AObuchow:
@mickaelistria I recreated the original patch and it's test (as well as `testPropertyCompletion` and `testParentPropertyCompletion`) pass when run locally. 

If you think this is OK then I'll merge it so @akurtakov can test it out (from m2e) as a user. However, there's no real rush, if you want to wait to run the test locally (or for tests to be running on CI) that's ok.

PR #29, opened 2020-04-23
Stop participants from processing non-pom files

### LINKED ISSUE #27: Lemminx Maven erroneously process non-pom.xml files
Severity: **** (blocker for first release)

Open any XML file with lemminx-maven available, and there will be pom error reported on the file.
It looks like all the "participants" have to add
```java
if (!MavenPlugin.match(...)) {
  return;
}
```

AObuchow:
Fix #27

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
@@ -101,6 +101,10 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 	@Override
 	public void onTagOpen(ICompletionRequest request, ICompletionResponse response)
 			throws Exception {
+		if (!MavenPlugin.match(request.getXMLDocument())) {
+			  return;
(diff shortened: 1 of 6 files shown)


AObuchow:
Currently, this patch breaks all the tests because most test xml files aren't called "pom.xml"
To fix this, I see two options:
1. Rename all of the test xml files to "pom.xml" and put them in separate folders to identify them correctly.
2. Add a system property which would allow the non-pom.xml files to be processed during tests.

IMO solution 2. is preferable (less work to fix existing tests). What do you think @mickaelistria?

mickaelistria:
I suggest we just change a bit the match condition so that `pom*.xml` and `*pom.xml` are mached too.

AObuchow:
@mickaelistria That would work too. It may lead to some very rare & unlucky cases where a non-pom file is processed, but as this issue seems to be a Lemminx upstream issue at its core (IMO), I think that's OK. After all, this patch is a workaround to an upstream issue.

AObuchow:
@mickaelistria I made it so `pom*.xml` and `*pom.xml` are matched.

PR #34, opened 2020-04-28
Try some support for setting

mickaelistria:

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
@@ -26,13 +26,11 @@ import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.InvalidRepositoryException;
-import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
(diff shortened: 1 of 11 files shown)


gayanper:
Any reason we can’t just set the global settings file into mave request object and let maven do the parsing?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -62,16 +61,17 @@
 	private final Map<URI, Collection<ModelProblem>> problemCache;
 	private final PlexusContainer plexusContainer;
 
-	private MavenExecutionRequest mavenRequest;
+	private final MavenExecutionRequest mavenRequest;
 	MavenXpp3Reader mavenReader = new MavenXpp3Reader();
 	private DefaultRepositorySystemSession repositorySystemSession;
 	private ProjectBuilder projectBuilder;
-	private RepositorySystem repositorySystem;
+	//private RepositorySystem repositorySystem;
```
AObuchow, line 68: Remove

mickaelistria:
> Any reason we can’t just set the global settings file into mave request object and let maven do the parsing?

As far as I can see, the mavenRequest is just a bean. It doesn't contain any logic to parse the settings file.

gayanper:
This change actually fixes the problem of not processing the settings.xml. Once the system properly is set to configuration file everything works like a charm.

AObuchow:
Looks good to me, nice improvement & refactoring @mickaelistria !

Issue #8, opened 2020-04-03
Completion/validation for "constrained" configuration element


mickaelistria:
From https://github.com/AObuchow/lsp4xml-extensions-maven/issues/64

`<configuration>` items for which we know the type is constrained (boolean, integer, enum...) should receive the possible values in completion proposal, and some validation (in case the value doesn't contain a property).

Difficulty: ****
Value: **..

AObuchow:
This should now be more easily possible due to the work done in https://github.com/eclipse/lemminx-maven/commit/862dd9bdbd53e7b4303c867f1b09fc4e89de8f9a & https://github.com/eclipse/lemminx-maven/commit/e655dd76f31fed97578d29aa18546cb88f81a2f9

PR #41, opened 2020-04-30
WIP Validate plugin configuration element types

AObuchow:
Part of #8

NOT READY, do NOT merge.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
@@ -19,6 +19,7 @@ import org.eclipse.lsp4j.DiagnosticSeverity;
 public class PluginValidator {
 	private final MavenProjectCache cache;
 	private final MavenPluginManager pluginManager;
+	private static final String FALSE = "false";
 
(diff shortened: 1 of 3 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -84,12 +85,105 @@ private static Diagnostic internalValidateConfiguration(DiagnosticRequest diagno
...
+		}
+		
+		DOMNode childNode = node.getChild(0);
+		if (childNode.isComment()) {
+			return;
+		}
+		
+		String value = childNode.getNodeValue();
+		switch (type) {
+		case "java.lang.Integer":
```
AObuchow, line 162: The limitation is that the type might be a custom interface defined in the MOJO/plugin, such is the case for [maven-enforcer-plugin](https://github.com/apache/maven-enforcer/blob/master/enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/EnforcerRule.java)

AObuchow:
There's also a regression that occured, needs to be fixed.

AObuchow:
Closing as we now have a better way to do this using the `MojoParameter` class

PR #44, opened 2020-05-01
Removed temp working file creating in the parent directory.

gayanper:
Change-Id: Ibdcf2b5722d889ba56b637136f6d8a2c86cbd71b
Signed-off-by: gayanper <gayanper@gmail.com>

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
@@ -10,12 +10,9 @@ package org.eclipse.lemminx.maven;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
-import java.nio.file.Files;
-import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -154,7 +151,8 @@ public class MavenProjectCache {
 			}, request);
 			problems.addAll(buildResult.getProblems());
 			if (buildResult.getProject() != null) {
-				// setFile should ideally be invoked during project build, but related methods to pass modelSource and pomFile are private
+				// setFile should ideally be invoked during project build, but related methods
+				// to pass modelSource and pomFile are private
 				buildResult.getProject().setFile(new File(uri));
 				projectCache.put(uri, buildResult.getProject());
 				projectParsedListeners.forEach(listener -> listener.accept(buildResult.getProject()));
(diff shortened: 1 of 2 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -170,9 +167,7 @@ public InputStream getInputStream() throws IOException {
 					File workingCopy = null;
 					try {
 						File file = new File(uri);
-						workingCopy = File.createTempFile("workingCopy", '.' + file.getName(), file.getParentFile());
-						Files.write(workingCopy.toPath(), document.getText().getBytes(), StandardOpenOption.CREATE);
-						Model model = mavenReader.read(new FileReader(workingCopy));
+						Model model = mavenReader.read(new ByteArrayInputStream(document.getText().getBytes()));
```
mickaelistria, line 170: This stream needs to be closed, or to be part of a try-with-resources block
```java
File file = new File(uri);
try (InputStream documentStream = new ByteArrayInputStream(document.getText().getBytes()) {
  Model model = mavenReader.read(documentStream);
  ...
}
```

AObuchow:
@gayanper I think this PR neesd to be rebased since the master branch was just recently updated.

mickaelistria:
FWIW, you don't need a Change-Id on GitHub, it's only used by Gerrit to show various versions of a commit. On GitHub, you can just amend and `git push --force` to your fork's PR branch and GitHub updates the PR automatically.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
gayanper: I thought about it is as well. But then again the Javadoc in ByteArrayInputStream doesn't hold any resources which need closing, it just backed by a byte array. Even the Javadoc says the close operation of ByteArrayInputStream does nothing. It's basically an empty method. So do we really need to complex the code here with the try with resource ?

mickaelistria: I'm afraid some static analysis engines or others can be confused because it's a Closeable and everyone/thing expects Closeable to be closed even if the Javadoc says it's not necessary for a given specific case.
The try-with-resources doesn't make code more complex actually, it extract the stream as a variable and wraps it in a try(...). IMO, it's better to make code seems more correct to automated or human readers, for the very small difference it makes in readability.

gayanper: agree I will fix it

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -167,24 +165,25 @@ public InputStream getInputStream() throws IOException {
 					File workingCopy = null;
 					try {
```
mickaelistria, line 166: you should get rid of this try(...) by the way.

gayanper: Done and also I clean up the remnant workingCopy file reference as well.

mickaelistria:
+1, looks good.

AObuchow:
Looks good to me, great first patch @gayanper! Glad to have you as a contributor.
Thanks @mickaelistria for the review!

PR #52, opened 2020-05-11
Add system properties to ProjectBuildRequest

mickaelistria:
That allows to properly resolve profiles and other stuff relying on
system properties.

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -31,7 +31,7 @@
 		<dependency>
 			<groupId>org.eclipse.lemminx</groupId>
 			<artifactId>org.eclipse.lemminx</artifactId>
-			<version>0.11.0</version>
+			<version>0.11.1</version>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.maven</groupId>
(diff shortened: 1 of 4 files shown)


PyvesB:
Thanks for picking this up. Out of interest, how do system properties related to BOM dependencies?

mickaelistria:
When writing the test case, I saw there were actually 2 errors, one being about "Failed to determine Java version for profile...". So I looked for this message in Maven code base (I have it locally, it's pretty useful to have it here when working on LemMinX-Maven or Maven) and found it's happening when the "java.version" system property isn't available in the projectBuildingRequest. So I just forwarding existing properties and saw the test pass.

AObuchow:
Although I dislike having to add software.amazon.awssdk:bom as a dependency to the build just for test's, it's the current state of lemminx-maven so I'll deal with it ;) This week I intend to look into maven dependency & plugin downloading so that Lemminx-Maven can download pom dependencies if they don't exist locally. Hopefully, that will remove the need for this new dependency.

AObuchow:
Other than that, looks good to me. Glad to see the fix was relatively simple. Great patch @mickaelistria !

PR #54, opened 2020-05-11
Fix false positives from PluginValidator

### LINKED ISSUE #48: Plugin configuration validation gives false-positives
There are a number of corner-cases that the [`PluginValidator`](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java) is missing. Test's should be added for all these cases.

In my testing, the cases which provide a false positive are:
1. When a plugin's groupId is missing. This gets resolved when the groupId is added. 
Example test case: 
```XML
			<plugin>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.8.1</version>
				<configuration>
					<source>1.8</source>
					<target>1.8</target>
				</configuration>
			</plugin>
```
**Note:** the warning does not appear when only the groupId & artifactId is present (for this test case)

2. When certain goals are present (in this example, the `enforce` goal gives a warning, but not the plugin configuration).
Example test case: 
```XML
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-enforcer-plugin</artifactId>
				<version>3.0.0-M3</version>
				<executions>
					<execution>
						<id>enforce-maven</id>
						<goals>
							<goal>enforce</goal>
						</goals>
						<configuration>
							<rules>
								<requireMavenVersion>
									<version>3.0.5</version>
								</requireMavenVersion>
							</rules>
						</configuration>
					</execution>
				</executions>
			</plugin>
```
**Note:** The goal which is giving a warning is giving it's description on hover, which means it was correctly resolved from the plugin descriptor. 
image

3. When a plugin's pom is missing from the `.m2/repository/` folder. This might not be the actual cause, but in the below example case, maven-shade-plugin `3.2.2` & `3.2.1` give plugin configuration errors and aren't able to resolve plugin goals (no goal description on hover). The error is resolved when version `3.2.3` or `2.3` is used. Below is a comparison of the contents of `3.2.2` & `3.2.3`:
```bash
[aobuchow@localhost 3.2.2]$ pwd
/home/aobuchow/.m2/repository/org/apache/maven/plugins/maven-shade-plugin/3.2.2
[aobuchow@localhost 3.2.2]$ ls -la
total 148
drwxr-xr-x. 2 aobuchow aobuchow   4096 May  6 12:28 .
drwxrwxr-x. 6 aobuchow aobuchow   4096 May  6 14:33 ..
-rw-r--r--. 1 aobuchow aobuchow     95 May  6 14:32 m2e-lastUpdated.properties
-rw-r--r--. 1 aobuchow aobuchow 130877 May  6 12:28 maven-shade-plugin-3.2.2.jar
-rw-r--r--. 1 aobuchow aobuchow     40 May  6 12:28 maven-shade-plugin-3.2.2.jar.sha1
-rw-r--r--. 1 aobuchow aobuchow    178 May  6 12:28 _remote.repositories
```
```bash
[aobuchow@localhost 3.2.3]$ pwd
/home/aobuchow/.m2/repository/org/apache/maven/plugins/maven-shade-plugin/3.2.3
[aobuchow@localhost 3.2.3]$ ls -la
total 168
drwxr-xr-x. 2 aobuchow aobuchow   4096 May  6 14:33 .
drwxrwxr-x. 6 aobuchow aobuchow   4096 May  6 14:33 ..
-rw-r--r--. 1 aobuchow aobuchow     95 May  6 14:34 m2e-lastUpdated.properties
-rw-r--r--. 1 aobuchow aobuchow 132252 May  6 14:33 maven-shade-plugin-3.2.3.jar
-rw-r--r--. 1 aobuchow aobuchow     40 May  6 14:33 maven-shade-plugin-3.2.3.jar.sha1
-rw-r--r--. 1 aobuchow aobuchow  10420 May  6 14:33 maven-shade-plugin-3.2.3.pom
-rw-r--r--. 1 aobuchow aobuchow     40 May  6 14:33 maven-shade-plugin-3.2.3.pom.sha1
-rw-r--r--. 1 aobuchow aobuchow    216 May  6 14:33 _remote.repositories
```
To briefly summarise this case: the plugin descriptor isin't being resolved for `maven-shade-plugin 3.2.2` (error on goals & configuration), but is being resolved for `maven-shade-plugin 3.2.3`.
Example test case:
```XML
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-shade-plugin</artifactId>
				<version>3.2.2</version>
				<executions>
					<execution>
						<phase>package</phase>
						<goals>
							<goal>shade</goal>
						</goals>
						<configuration>
							<transformers>
								<transformer
									implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
									<mainClass>org.eclipse.dash.licenses.cli.Main</mainClass>
								</transformer>
							</transformers>
							<createDependencyReducedPom>
								false
							</createDependencyReducedPom>
						</configuration>
					</execution>
				</executions>
			</plugin>
```

AObuchow:
Fix #48

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -98,6 +98,26 @@
 			<type>pom</type>
 			<scope>test</scope>
 		</dependency>
+	<!-- This is a workaround for CI because it looks like CI cannot connect 
+		to Central during tests -->
+	<!-- Use in LocalPluginTest#testPluginDiagnosticMissingPlugin -->
+	<dependency>
+		<groupId>org.apache.maven.plugins</groupId>
+		<artifactId>maven-enforcer-plugin</artifactId>
+		<version>3.0.0-M3</version>
+		<type>jar</type>
+		<scope>test</scope>
+	</dependency>
+	<!-- This is a workaround for CI because it looks like CI cannot connect 
+		to Central during tests -->
+	<!-- Use in LocalPluginTest#testPluginDiagnosticMissingPlugin -->
+	<dependency>
+		<groupId>org.apache.maven.plugins</groupId>
+		<artifactId>maven-shade-plugin</artifactId>
+		<version>3.2.2</version>
+		<type>pom</type>
+		<scope>test</scope>
+	</dependency>
 	</dependencies>
 
 	<build>
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
@@ -8,10 +8,17 @@
  *******************************************************************************/
 package org.eclipse.lemminx.maven;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
+import org.apache.maven.plugin.InvalidPluginDescriptorException;
 import org.apache.maven.plugin.MavenPluginManager;
+import org.apache.maven.plugin.PluginDescriptorParsingException;
+import org.apache.maven.plugin.PluginResolutionException;
+import org.apache.maven.plugin.descriptor.MojoDescriptor;
 import org.apache.maven.plugin.descriptor.Parameter;
+import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lsp4j.Diagnostic;
@@ -22,10 +29,38 @@ public class PluginValidator {
 	private final MavenPluginManager pluginManager;
 	private final RepositorySystemSession repoSession;
 
-	public PluginValidator(MavenProjectCache cache, RepositorySystemSession repoSession, MavenPluginManager pluginManager) {
+	public PluginValidator(MavenProjectCache cache, RepositorySystemSession repoSession,
+			MavenPluginManager pluginManager) {
 		this.cache = cache;
 		this.repoSession = repoSession;
 		this.pluginManager = pluginManager;
+	}
+	
+	public boolean validatePluginResolution(DiagnosticRequest diagnosticRequest) {
+		try {
+			MavenPluginUtils.getContainingPluginDescriptor(diagnosticRequest, cache, repoSession,
+					pluginManager);
+		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
+			//Add artifactId diagnostic
+			e.printStackTrace();
+			String errorMessage = e.getMessage();
+			DOMNode pluginNode = DOMUtils.findClosestParentNode(diagnosticRequest, "plugin");
+			Optional<DOMNode> artifactNode = pluginNode.getChildren().stream().filter(node -> node.getLocalName().equals("artifactId")).findAny();
+			artifactNode.ifPresent(node -> {
+				DiagnosticRequest artifactDiagnosticReq = new DiagnosticRequest(node,
+						diagnosticRequest.getXMLDocument(), diagnosticRequest.getDiagnostics());
+				Diagnostic unresolvablePlugin = new Diagnostic(artifactDiagnosticReq.getRange(),
+						"Plugin could not be resolved. Ensure the plugin's groupId, artifactId and version are present." + MavenPluginUtils.LINE_BREAK + "Additional information: "
+								+ errorMessage, DiagnosticSeverity.Warning,
+				artifactDiagnosticReq.getXMLDocument().getDocumentURI(), "XML");
+				List<Diagnostic> diagnostics = diagnosticRequest.getDiagnostics();
+				if (!diagnostics.contains(unresolvablePlugin)) {
+					diagnosticRequest.getDiagnostics().add(unresolvablePlugin);					
+				}
+			});
+			return false;
+		}
+		return true;
 	}
 
 	public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
@@ -33,9 +68,23 @@ public class PluginValidator {
 		if (node == null) {
 			return null;
 		}
+		if (!validatePluginResolution(diagnosticRequest)) {
+			return null;
+		}
+		
+		
+		List<Parameter> parameters = new ArrayList<>();
+		try {
+			parameters = MavenPluginUtils.collectPluginConfigurationParameters(diagnosticRequest, cache, repoSession,
+					pluginManager);
+		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
+			e.printStackTrace();
+			// A diagnostic was already added in validatePluginResolution()
+		}
+		if (parameters.isEmpty()) {
+			return null;
+		}
 		if (node.isElement() && node.hasChildNodes()) {
-			List<Parameter> parameters = MavenPluginUtils.collectPluginConfigurationParameters(diagnosticRequest, cache,
-					repoSession, pluginManager);
 			for (DOMNode childNode : node.getChildren()) {
 				DiagnosticRequest childDiagnosticReq = new DiagnosticRequest(childNode,
 						diagnosticRequest.getXMLDocument(), diagnosticRequest.getDiagnostics());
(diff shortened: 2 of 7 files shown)


AObuchow:
This is WIP. It still needs to fix Case 3 & be refactored + cleaned up as much as possible.

AObuchow:
There's also a regression that occurred from this patch, needs to be fixed.

AObuchow:
I looked into the regression for `LocalPluginTest.testPluginGoalDiagnostics`. The erroneous diagnostic that was now being generated was: 
```
Diagnostic [
  range = Range [
    start = Position [
      line = 24
      character = 21
    ]
    end = Position [
      line = 26
      character = 6
    ]
  ]
  severity = Warning
  code = "XML"
  source = "file:/home/aobuchow/git/lemminx-maven/lemminx-maven/target/test-classes/pom-plugin-goal-diagnostic.xml"
  message = "Plugin parameters could not be resolved. Ensure the plugin's groupId, artifactId and version are present."
  relatedInformation = null
]
```

Checking [if the plugin's configuration parameters is empty](https://github.com/eclipse/lemminx-maven/pull/54/files#diff-75b35d7fdfb0eab29a93b39bd6847f17R43) is too vague to give a correct diagnostic. In this case, the goal has no configuration parameters, so any parameter given within it's configuration tags will give a false-positive diagnostic.

mickaelistria:
I think you need to the right maven-enforcer-plugin versiond described in pom.xml (like the aws stuff) for tests to pass on CI.

AObuchow:
That’s true. I’m going to aim to get the regression tests passing locally and then see if I can implement maven dependency downloading to fix the tests on CI in a more elegant manner. Otherwise I’ll resort to the workaround.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -46,65 +48,105 @@ public void tearDown() throws InterruptedException, ExecutionException {
...
+		}, new XMLValidationSettings()).stream().map(Diagnostic::getMessage)
+				.anyMatch(message -> message.contains("Invalid goal for this plugin")));
+		List<Diagnostic> diags = languageService.doDiagnostics(document, () -> {
+		}, new XMLValidationSettings());
+		System.out.println(diags);
+		assertTrue(languageService.doDiagnostics(document, () -> {
+		}, new XMLValidationSettings()).size() == 2);
+	}
+
+	// Tests regressions related to
```
AObuchow, line 136: This file was accidentally formatted... the real change to LocalPluginTest is here.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -46,65 +48,105 @@ public void tearDown() throws InterruptedException, ExecutionException {
...
+	// https://github.com/eclipse/lemminx-maven/issues/48
+	@Test
+	public void testPluginDiagnosticsIssue48()
```
AObuchow, line 139: Maybe the test should have a better name? I do like the idea of associating a test with a GitHub issue for tracking reasons however.

lemminx-maven/src/test/resources/pom-plugin-configuration-diagnostic-issue48.xml
```diff
@@ -0,0 +1,77 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
AObuchow, line 1: Rename this file to `pom-plugin-diagnostic-issue48.xml` as it deals with a bug in the goal & configuration diagnostics.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -54,28 +74,38 @@ public Diagnostic validateGoal(DiagnosticRequest diagnosticRequest) {
...
+
+		String goal = node.getChild(0).getNodeValue();
+		for (MojoDescriptor mojo : pluginDescriptor.getMojos()) {
+			if (!goal.isEmpty() && goal.equals(mojo.getGoal())) {
 				return null;
 			}
 		}
-		return new Diagnostic(diagnosticReq.getRange(), "Invalid goal for this plugin", DiagnosticSeverity.Warning,
-				diagnosticReq.getXMLDocument().getDocumentURI(), "XML");
+		System.out.println("goa: " + goal);
```
AObuchow, line 105: Remove

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
mickaelistria: I prefer explicit names that don't require to look at some other link to get an idea of what they're testing.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -54,28 +74,38 @@ public Diagnostic validateGoal(DiagnosticRequest diagnosticRequest) {
...
+			try {
+				pluginDescriptor = MavenPluginUtils.getContainingPluginDescriptor(diagnosticRequest, cache, repoSession,
+						pluginManager);
+				Diagnostic diag = internalValidateGoal(diagnosticRequest, pluginDescriptor);
+				if (diag != null) {
+					diagnosticRequest.getDiagnostics().add(diag);
+				}
+			} catch (PluginResolutionException | PluginDescriptorParsingException
+					| InvalidPluginDescriptorException e) {
+				e.printStackTrace();
```
AObuchow, line 87: @mickaelistria do you think goals should also give a diagnostic if the plugin wasn't resolvable? Something similar to [line 51](https://github.com/eclipse/lemminx-maven/pull/54/files#diff-75b35d7fdfb0eab29a93b39bd6847f17R51)

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
AObuchow: Ok sounds good, maybe it should be something like `testPluginDiagnosticEdgeCases` or `testPluginDiagnosticFalsePositives`?

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -46,65 +48,105 @@ public void tearDown() throws InterruptedException, ExecutionException {
...
+			throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-plugin-configuration-diagnostic-issue48.xml", languageService);
+		System.out.println(languageService.doDiagnostics(document, () -> {
```
AObuchow, line 142: Remove (although being able to see the diagnostic is kind of useful... maybe further test suite improvements could collect the diagnostic into the test results)

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
mickaelistria: More something like `testPluginDiagnosticMissingPluginConfiguration`

lemminx-maven/pom.xml
```diff
@@ -102,6 +102,22 @@
 
 	<build>
 		<plugins>
+	<!-- This is a workaround for CI because it looks like CI cannot connect
```
AObuchow, line 105: Seems like CI is still failing while including these plugins in the build. Maybe they should be included in the project's dependencies instead?

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
AObuchow: +1

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
mickaelistria: I think if we cannot get plugin info, we should only show one diagnostic for goal/configuration.

AObuchow: So rather than have a diagnostic show up under goal & configuration, only make the diagnostic show on one of them? 
Ideally, I should try and make the plugin's GAV tags get this "unresolvable plugin" warning diagnostic.

lemminx-maven/pom.xml
```diff
@@ -98,10 +98,46 @@
...
+		<scope>test</scope>
+	</dependency>
 	</dependencies>
 
 	<build>
 		<plugins>
+	<!-- This is a workaround for CI because it looks like CI cannot connect 
+		to Central during tests -->
+	<!-- Use in LocalPluginTest#testPluginDiagnosticMissingPlugin -->
+	<plugin>
```
AObuchow, line 128: Try removing these and see if CI test still passes. I believe they should.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
AObuchow: I ended up making a plugin's <artifactId> tag get a diagnostic if a plugin could not be resolved. In this case, all configuration/goal diagnostics are disabled.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -22,20 +29,62 @@
...
+		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
+			//Add artifactId diagnostic
+			e.printStackTrace();
+			String errorMessage = e.getMessage();
+			DOMNode pluginNode = DOMUtils.findClosestParentNode(diagnosticRequest, "plugin");
+			Optional<DOMNode> artifactNode = pluginNode.getChildren().stream().filter(node -> node.getLocalName().equals("artifactId")).findAny();
+			artifactNode.ifPresent(node -> {
+				DiagnosticRequest artifactDiagnosticReq = new DiagnosticRequest(node,
+						diagnosticRequest.getXMLDocument(), diagnosticRequest.getDiagnostics());
+				Diagnostic unresolvablePlugin = new Diagnostic(artifactDiagnosticReq.getRange(),
```
AObuchow, line 52: A diagnosticRequest should be able to generate a diagnostic, given an error message and severity. Refactor this.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -107,4 +107,25 @@ public void testPluginGoalDiagnostics() throws IOException, InterruptedException
...
+			DOMDocument document = createDOMDocument("/pom-plugin-diagnostic-missing-plugin.xml", languageService);
+			// One diagnostic should be from a missing groupID (incomplete GAV), another should be
+			// from unresolvable plugin (maven-shade-plugin:3.2.2)
+			assertTrue(languageService.doDiagnostics(document, () -> {
+			}, new XMLValidationSettings()).stream().map(Diagnostic::getMessage).anyMatch(message -> message.contains(
+					"Plugin org.apache.maven.plugins:maven-shade-plugin:3.2.2 or one of its dependencies could not be resolved")));
+			assertTrue(languageService.doDiagnostics(document, () -> {
+			}, new XMLValidationSettings()).stream().map(Diagnostic::getMessage).anyMatch(message -> message.contains(
+					"Incomplete GAV")));
+			System.out.println(languageService.doDiagnostics(document, () -> {
```
AObuchow, line 126: Remove

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
AObuchow: Refactor this in a (very near-)future PR.

AObuchow:
Merging this so it gets into M3, further improvements can be added in future PR's.

PR #55, opened 2020-05-14
Remove duplicate plugin configuration completions

### LINKED ISSUE #53: Duplicate plugin configuration completion's
When triggering a completion within a `<configuration>...</configuration>` tag, it seems that duplicate completion's can exist. The description of these duplicate completions seems exactly the same, so we can probably filter them out by comparing existing completion descriptions (for plugin configuration).

See below for a comparison of the completion descriptions:
image
image

AObuchow:
Fix #53

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
@@ -24,10 +24,12 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.apache.maven.Maven;
@@ -248,6 +250,8 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 		case "configuration":
 			MavenPluginUtils.collectPluginConfigurationParameters(request, cache, repoSession, pluginManager).stream()
 					.map(parameter -> toTag(parameter.getName(), MavenPluginUtils.getMarkupDescription(parameter), request))
+					.filter(distinctByKey(
+							completionItem -> ((CompletionItem) completionItem).getDocumentation().getLeft()))
 					.forEach(response::addCompletionItem);
 			break;
 //		case "relativePath":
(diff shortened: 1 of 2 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
```diff
@@ -110,7 +112,8 @@ public void onTagOpen(ICompletionRequest request, ICompletionResponse response)
 		
 		if ("configuration".equals(request.getParentElement().getLocalName())) {
 			MavenPluginUtils.collectPluginConfigurationParameters(request, cache, repoSession, pluginManager).stream()
-					.map(parameter -> toTag(parameter.getName(), MavenPluginUtils.getMarkupDescription(parameter), request))
+					.map(parameter -> toTag(parameter.getName(), MavenPluginUtils.getMarkupDescription(parameter), request)).filter(distinctByKey(
+							completionItem -> ((CompletionItem) completionItem).getDocumentation().getLeft()))
```
AObuchow, line 116: @mickaelistria do you think there's a more elegant way to filter out duplicates by comparing a specific property?
I kind of dislike the fact I had to cast the `completionItem` to a `CompletionItem` (because the `distinctByKey` expects a function which takes in an `Object`).

mickaelistria: I don't have a better idea at the moment.

AObuchow: > I don't have a better idea at the moment.

No worries

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
```diff
@@ -248,6 +250,8 @@ public void onXMLContent(ICompletionRequest request, ICompletionResponse respons
 		case "configuration":
 			MavenPluginUtils.collectPluginConfigurationParameters(request, cache, repoSession, pluginManager).stream()
 					.map(parameter -> toTag(parameter.getName(), MavenPluginUtils.getMarkupDescription(parameter), request))
+					.filter(distinctByKey(
```
AObuchow, line 253: This filter logic should probably be moved to `MavenPluginUtils.collectPluginConfigurationParameters` to prevent other future duplicate configuration parameters from popping up elsewhere.

AObuchow: The duplicates actually aren't occurring in MavenPluginUtils.collectPluginConfigurationParameters.

AObuchow:
@mickaelistria @angelozerr I'm still a bit confused from the original cause of this bug.
There are no duplicates coming from `MavenPluginUtils.collectPluginConfigurationParameters`.

My current suspicion is that the same completion is being added twice, once in [`onTagOpen`](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java#L114) and once in[ `onXMLContent`](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java#L122). 

There's no way to determine if a completion was already added to the completion response with Lemminx's extension API. Perhaps Lemminx should filter duplicate completions from the same extension?

However... if the completion being added to the response twice is the root issue, that doesn't explain why my fix solves the issue.

AObuchow:
Despite the root cause of this bug still being unknown, I'm going to merge this patch as it fixes the issue. When the root cause is found, we should revert this patch and address the root cause.

PR #62, opened 2020-05-19
Support for plugin version set in pluginManagement

mickaelistria:


AObuchow:
Looks good to me, thanks for adding support for this case!

PR #66, opened 2020-05-20
Add click-to-definition for maven properties

### LINKED ISSUE #3: Ctrl+Click (definition) on ${...} should link to property definition
From https://github.com/AObuchow/lsp4xml-extensions-maven/issues/91

Ctrl+Click (definition) on ${...} should link to property definition

Difficulty: **..
Value (frequency*time saved): ***.

AObuchow:
Fix #3

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
@@ -8,9 +8,12 @@
  *******************************************************************************/
 package org.eclipse.lemminx.maven;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.Optional;
 
 import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lemminx.services.extensions.ICompletionRequest;
@@ -33,6 +36,26 @@ public class DOMUtils {
 
 		if (localName.equals(pluginNode.getLocalName())) {
 			return pluginNode;
+		}
+		return null;
+	}
+	
+	// TODO: use optional?
+	public static DOMNode findNodeByLocalName(final DOMDocument document, final String localName) {
+		Deque<DOMNode> nodes = new ArrayDeque<>();
+		for (DOMNode node : document.getChildren()) {
+			nodes.push(node);
+		}
+		while (!nodes.isEmpty()) {
+			DOMNode node = nodes.pop();
+				if (node.getLocalName() != null && node.getLocalName().equals(localName)) {
+					return node;
+				}
+			if (node.hasChildNodes()) {
+				for (DOMNode childNode : node.getChildren()) {
+					nodes.push(childNode);
+				}
+			}
 		}
 		return null;
 	}
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
@@ -16,6 +16,7 @@ import java.util.Optional;
 import org.apache.maven.Maven;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.project.MavenProject;
+import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lemminx.maven.searcher.LocalRepositorySearcher;
@@ -44,11 +45,18 @@ public class MavenDefinitionParticipant implements IDefinitionParticipant {
 		}
 		
 		File currentFolder = new File(URI.create(request.getXMLDocument().getTextDocument().getUri())).getParentFile();
+		
+		LocationLink propertyLocation = findMavenPropertyLocation(request);
+		if (propertyLocation != null) {
+			locations.add(propertyLocation);
+			return;
+		}
+		
 		DOMElement element = findInterestingElement(request.getNode());
 		if (element.getLocalName().equals("module")) {
-			File subModuleFile = new File(currentFolder, element.getFirstChild().getTextContent() + "/" + Maven.POMv4);
+			File subModuleFile = new File(currentFolder, element.getFirstChild().getTextContent() + File.separator + Maven.POMv4);
 			if (subModuleFile.isFile()) {
-				locations.add(toLocation(subModuleFile, element));
+				locations.add(toLocationNoRange(subModuleFile, element));
 			}
 			return;
 		}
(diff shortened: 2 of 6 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
```diff
@@ -62,27 +75,53 @@ public void findDefinition(IDefinitionRequest request, List<LocationLink> locati
...
 			}
 			return;
 		}
 	}
 
+	private LocationLink findMavenPropertyLocation(String mavenProperty, IDefinitionRequest request) {
+		DOMDocument xmlDocument = request.getXMLDocument();
+		MavenProject project = cache.getLastSuccessfulMavenProject(xmlDocument);
+		MavenProject childProj = project;
+		while (project != null && project.getProperties().containsKey(mavenProperty)) {
```
AObuchow, line 103: There needs to be a check for the case that the property wasn't defined in the project, ie. the property is invalid

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
```diff
@@ -62,27 +75,53 @@ public void findDefinition(IDefinitionRequest request, List<LocationLink> locati
...
+			childProj = project;
+			project = project.getParent();
+		}
+		
+		DOMNode propertyDeclaration = null;
+		// The substring is an ugly hack to get the path of the xmlDocument
+		if (childProj.getFile().toURI().getPath().equals(xmlDocument.getDocumentURI().substring(5))) {
```
AObuchow, line 110: @mickaelistria any suggestion on how to better compare that the xml document & maven project's file are equal? I had to remove the `"file:"` part of the documentURI string.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
```diff
@@ -62,27 +75,53 @@ public void findDefinition(IDefinitionRequest request, List<LocationLink> locati
...
+			// Property is defined in the same file as the request
+			propertyDeclaration = DOMUtils.findNodeInDocument(xmlDocument, mavenProperty);
+		} else {
+			DOMDocument propertyDeclaringDocument = org.eclipse.lemminx.utils.DOMUtils.loadDocument(childProj.getFile().toURI().toString(),
+					request.getNode().getOwnerDocument().getResolverExtensionManager());
+			propertyDeclaration = DOMUtils.findNodeInDocument(propertyDeclaringDocument, mavenProperty);
+		}
+		
+		if (propertyDeclaration == null) {
```
AObuchow, line 119: Technically, this should cover the case where the property doesn't exist, since the pom being searched for the property's declaration will be missing the propertyDeclaration node.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
```diff
@@ -99,11 +138,18 @@ private File getArtifactLocation(Dependency dependency) {
...
 
-	private LocationLink toLocation(File target, DOMElement element) {
+	private LocationLink toLocationNoRange(File target, DOMNode element) {
 		Range dumbRange = new Range(new Position(0, 0), new Position(0, 0));
 		LocationLink link = new LocationLink(target.toURI().toString(), dumbRange, dumbRange, XMLPositionUtility.createRange(element));
 		return link;
 	}
+	
+	private LocationLink toLocation(File target, DOMNode element, IDefinitionRequest request) {
+		Range dumbRange = XMLPositionUtility.createRange(((DOMElement) element).getStartTagCloseOffset() + 1,
```
AObuchow, line 148: Don't call this dumbRange as it's dynamic here.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -238,7 +238,7 @@ public String onText(IHoverRequest request) throws Exception {
 	}
 
 
-	public String getMavenPropertyInHover(IPositionRequest request) {
+	public static String getMavenPropertyInHover(IPositionRequest request) {
```
AObuchow, line 241: rename to `getMavenPropertyInRequest`

AObuchow:
it'd be nice to also add test coverage for other definitions that are supported, although this could be for a future patch

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -141,6 +141,24 @@ public void testPropertyHover() throws IOException, InterruptedException, Execut
...
  		assertNull(hover);
 	}
+	
+	@Test
+ 	public void testPropertyDefinitionSameDocument() throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-with-properties-for-definition.xml", languageService);
+		Position pos = new Position(14, 22);
+		List<? extends LocationLink> actual = languageService.findDefinition(document, pos, () -> {
+		});
+		assertTrue(actual.stream().anyMatch(link -> link.getTargetUri().endsWith("pom-with-properties-for-definition.xml")));
```
AObuchow, line 151: Note that we aren't testing that the target location is correct, maybe we should?...

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
mickaelistria: As mentioned later, this is actually check before returning, so no strong need for an extra check here.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
mickaelistria: would `childProj.getFile().toURI(),toString().equals(xmlDocument.getDocumentURI))` work?

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
mickaelistria: If it's not too difficult, it would be good to also add the check for it now to detect possible regressions in the future. However, it's not a strong requirement.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
AObuchow: agreed :)

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
AObuchow: Works perfectly :)!

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
AObuchow: Done :)

AObuchow:
@mickaelistria I think this is ready for a merge. 
My only issue with it is that it doesn't test (and might not support) the case where a Maven property is coming from a non-local pom.

Is that a real scenario? I'm uncertain whether inheriting a pom guarantees that the parent pom will be available locally.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
```diff
@@ -36,6 +39,26 @@ public static DOMNode findClosestParentNode(final IPositionRequest request, fina
 		}
 		return null;
 	}
+	
+	// TODO: use optional?
+	public static DOMNode findNodeInDocument(final DOMDocument document, final String localName) {
```
AObuchow, line 44: @angelozerr is there a better way to do this? Maybe something in the Lemminx API that I missed?

The function just searches all nodes in a document for a given node.

mickaelistria:
> My only issue with it is that it doesn't test (and might not support) the
> case where a Maven property is coming from a non-local pom. Is that a real
> scenario? I'm uncertain whether inheriting a pom guarantees that the parent
> pom will be available locally.
>
I think it's OK as building the MavenProject requires to resolve parent pom
and have them locally.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
angelozerr: No LemMinx doesn't provides this kind of search because it's too specifiic. You could use XPathManager for that, but as it's a simple search I will do that like you have done.

I will rename your method with `findNodeByLocalName`. What are you using a copy of nodes? Looping for nodes directly from DOM document should work too, no?

AObuchow: Good suggestion on the rename - going to do that :)

The reason I copy the nodes is so that the `while` condition evaluates to false until all nodes in the document have been checked. Future child nodes that get encountered will be added to the list of nodes to iterrate through. My technique might be redundant however :P Maybe there's a better way to do this.

AObuchow:
@mickaelistria okay awesome, I was hoping that would be the case.

AObuchow:
Merging as this seems good for now.

PR #67, opened 2020-05-20
Disable cache

mickaelistria:


AObuchow:
@mickaelistria I'm a bit divided on what to think of this finding. I'm glad you found out the issues that stem from using the MavenProjectCache (really good job there) but am also a bit disappointed the project cache causes issues (as it's a nice utility).

Do you see us abandoning the project cache, or improving it somehow to fix the issues it causes?

mickaelistria:
The Maven project cache is really not usable for case of development tools where pom can actually change. it's only focusing on running `mvn` command-line not on using Maven as API; m2e also disables this as well. I don't think there is other alternative available at the moment, as this cache is mostly internal and cannot easily be overriden AFAIK.
IMO, this would need to be fixed in Maven, allowing the cache to automatically "uncache" content when the underlying pom did change. Are you willing to open this issue to Maven tracker?

AObuchow:
Ok, thanks for the clarification.
Yes I'll open the issue in the Maven tracker - will get that setup as soon as I finish cleaning up my click-to-definition patch (nearly done).

AObuchow:
See https://issues.apache.org/jira/browse/MNG-6911 for the Maven upstream issue

mickaelistria:
@AObuchow Is this good enough or is it anything I should try to change before we merge?

AObuchow:
@mickaelistria looks good to me :) I was hoping we'd get a response upstream but this workaround is very good for the time being. I'm going to rerun the merge CI build to get tests to pass before the merge.

Just to make sure: you set resolve dependencies to true so that the parent will be resolved in the case that the parent pom's model hasn't been cached, right? (or something along those lines).

mickaelistria:
For the setResolveDepedencies(true), it's actuall not related to the issue.
But it should usually lead to better results, as it would download all referenced libraries locally, making them available for further hover and so on.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -16,7 +16,6 @@
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.Collections;
-import java.util.List;
```
AObuchow, line 19: Can you put back the import for `java.util.List`?
It's giving a compilation error on merge with the master branch:
```bash
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jenkins/agent/workspace/LemMinX-Maven_PR-67-merge/lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java:[152,17] cannot find symbol
  symbol:   class List
  location: class org.eclipse.lemminx.maven.test.SimpleModelTest
[ERROR] /home/jenkins/agent/workspace/LemMinX-Maven_PR-67-merge/lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java:[167,17] cannot find symbol
  symbol:   class List
  location: class org.eclipse.lemminx.maven.test.SimpleModelTest
```

AObuchow:
> For the setResolveDepedencies(true), it's actuall not related to the issue.
> But it should usually lead to better results, as it would download all referenced libraries locally, making them available for further hover and so on.

Ah okay that's awesome, seems like a lot of value for one extra line of code :)

mickaelistria:
I rebased my patch.

AObuchow:
Looks good & tests are passing, great patch @mickaelistria!

PR #73, opened 2020-05-26
Diagnostics refactor + missing but resolvable groupId fix

### LINKED ISSUE #48: Plugin configuration validation gives false-positives

### LINKED ISSUE #74: Check Maven Project/Model when plugin is unresolvable
In certain cases, a plugin's groupId or version tag may be missing in a given pom, but the plugin can still be resolved. In this case, the Maven Project or Model should be checked to see if it contains the plugin that is being resolved. 

Example test case:
```XML
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <groupId>demo</groupId>
  <artifactId>demo</artifactId>
  <build>
    <plugins>
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.8.1</version>
        <configuration>
          <release>14</release>
          <compilerArgs>
            <arg>--enable-preview</arg>
          </compilerArgs>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project>
```

Related m2e issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=563624

AObuchow:
Reenable  goal diagnostics

Test false goal diagnostics (part of #48)

If the groupId of a plugin is missing but its artifactId is present,
search the projects plugins for it in order to resolve the plugin (part
of #74)

Fix #48
Fix #74

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -115,8 +115,16 @@
 		<groupId>org.apache.maven.plugins</groupId>
 		<artifactId>maven-shade-plugin</artifactId>
 		<version>3.2.2</version>
-		<type>pom</type>
+		<type>test-jar</type>
 		<scope>test</scope>
+	</dependency>
+		<!-- This is a workaround for CI because it looks like CI cannot connect 
+		to Central during tests -->
+	<!-- Use in LocalPluginTest#testDiagnosticMissingGroupId -->
+	<dependency>
+		<groupId>org.apache.maven.plugins</groupId>
+		<artifactId>maven-compiler-plugin</artifactId>
+		<version>3.8.1</version>
 	</dependency>
 	</dependencies>
 
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DiagnosticRequest.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DiagnosticRequest.java
@@ -8,8 +8,6 @@
  *******************************************************************************/
 package org.eclipse.lemminx.maven;
 
-import java.util.List;
-
 import org.eclipse.lemminx.commons.BadLocationException;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
@@ -18,48 +16,42 @@ import org.eclipse.lemminx.dom.LineIndentInfo;
 import org.eclipse.lemminx.services.extensions.IPositionRequest;
 import org.eclipse.lemminx.utils.XMLPositionUtility;
 import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticSeverity;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 
 public class DiagnosticRequest implements IPositionRequest {
 	private DOMNode node;
 	private DOMDocument xmlDocument;
-	private List<Diagnostic> diagnostics;
-
-	public DiagnosticRequest(DOMNode node, DOMDocument xmlDocument, List<Diagnostic> diagnostics) {
-		this.setNode(node);
-		this.setXMLDocument(xmlDocument);
-		this.setDiagnostics(diagnostics);
+	
+	/**
+	 * A diagnosticRequest allows creation of diagnostics for a supplied node.
+	 * 
+	 * @param node        The node where diagnostics should be added.
+	 * @param xmlDocument The XMLDocument where the diagnostics will appear
+	 */
+	public DiagnosticRequest(DOMNode node, DOMDocument xmlDocument) {
+		this.node = node;
+		this.xmlDocument = xmlDocument;
 	}
-
+	
+	public Diagnostic createDiagnostic(String errorMessage, DiagnosticSeverity severity) {
+		return new Diagnostic(this.getRange(), errorMessage, severity, this.getXMLDocument().getDocumentURI(), "XML");
+	}
+	
+	private Range getRange() {
+		return XMLPositionUtility.createRange(((DOMElement) node).getStartTagCloseOffset() + 1,
+				((DOMElement) node).getEndTagOpenOffset(), xmlDocument);
+	}
+	
+	@Override
 	public DOMDocument getXMLDocument() {
 		return xmlDocument;
-	}
-
-	private void setXMLDocument(DOMDocument xmlDocument) {
-		this.xmlDocument = xmlDocument;
 	}
 
 	@Override
 	public DOMNode getNode() {
 		return node;
-	}
-
-	private void setNode(DOMNode node) {
-		this.node = node;
-	}
-
-	public List<Diagnostic> getDiagnostics() {
-		return diagnostics;
-	}
-
-	private void setDiagnostics(List<Diagnostic> diagnostics) {
-		this.diagnostics = diagnostics;
-	}
-
-	public Range getRange() {
-		return XMLPositionUtility.createRange(((DOMElement) node).getStartTagCloseOffset() + 1,
-				((DOMElement) node).getEndTagOpenOffset(), xmlDocument);
 	}
 
 	@Override
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDiagnosticParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDiagnosticParticipant.java
@@ -12,7 +12,10 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
@@ -50,8 +53,7 @@ public class MavenDiagnosticParticipant implements IDiagnosticsParticipant {
 		
 		projectCache.getProblemsFor(xmlDocument).stream().map(this::toDiagnostic).forEach(diagnostics::add);
 		DOMElement documentElement = xmlDocument.getDocumentElement();
-		HashMap<String, Function<DiagnosticRequest, Diagnostic>> tagDiagnostics = configureDiagnosticFunctions(
-				xmlDocument);
+		HashMap<String, Function<DiagnosticRequest, Optional<List<Diagnostic>>>> tagDiagnostics = configureDiagnosticFunctions();
 
 		Deque<DOMNode> nodes = new ArrayDeque<>();
 		for (DOMNode node : documentElement.getChildren()) {
@@ -59,20 +61,16 @@ public class MavenDiagnosticParticipant implements IDiagnosticsParticipant {
 		}
 		while (!nodes.isEmpty()) {
 			DOMNode node = nodes.pop();
-			for (String tagToValidate : tagDiagnostics.keySet()) {
-				if (node.getLocalName() != null && node.getLocalName().equals(tagToValidate)) {
-					Diagnostic diagnostic = null;
-					try {
-						diagnostic = tagDiagnostics.get(tagToValidate)
-								.apply(new DiagnosticRequest(node, xmlDocument, diagnostics));
-					} catch (Exception e) {
-						// TODO: Use plug-in error logger
-						e.printStackTrace();
-					}
-
-					if (diagnostic != null) {
-						diagnostics.add(diagnostic);
-					}
+			for (Entry<String, Function<DiagnosticRequest, Optional<List<Diagnostic>>>> entry : tagDiagnostics
+					.entrySet()) {
+				if (node.getLocalName() != null && node.getLocalName().equals(entry.getKey())) {
+					entry.getValue().apply(new DiagnosticRequest(node, xmlDocument))
+							.ifPresent(diagnosticList -> {
+								// Don't add a diagnostic if it already exists
+								diagnostics.addAll(
+										diagnosticList.stream().filter(diagnostic -> !diagnostics.contains(diagnostic))
+												.collect(Collectors.toList()));
+							});
 				}
 			}
 			if (node.hasChildNodes()) {
@@ -83,14 +81,15 @@ public class MavenDiagnosticParticipant implements IDiagnosticsParticipant {
 		}
 	}
 
-	private HashMap<String, Function<DiagnosticRequest, Diagnostic>> configureDiagnosticFunctions(
-			DOMDocument xmlDocument) {
+	private HashMap<String, Function<DiagnosticRequest, Optional<List<Diagnostic>>>> configureDiagnosticFunctions() {
 		PluginValidator pluginValidator = new PluginValidator(projectCache, repoSession, pluginManager);
-		Function<DiagnosticRequest, Diagnostic> validatePluginConfiguration = pluginValidator::validateConfiguration;
-		Function<DiagnosticRequest, Diagnostic> validatePluginGoal = pluginValidator::validateGoal;
 
-		HashMap<String, Function<DiagnosticRequest, Diagnostic>> tagDiagnostics = new HashMap<>();
+		Function<DiagnosticRequest, Optional<List<Diagnostic>>> validatePluginConfiguration = pluginValidator::validateConfiguration;
+		Function<DiagnosticRequest, Optional<List<Diagnostic>>> validatePluginGoal = pluginValidator::validateGoal;
+
+		HashMap<String, Function<DiagnosticRequest, Optional<List<Diagnostic>>>> tagDiagnostics = new HashMap<>();
 		tagDiagnostics.put("configuration", validatePluginConfiguration);
+		tagDiagnostics.put("goal", validatePluginGoal);
 		return tagDiagnostics;
 	}
 
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
@@ -10,10 +10,12 @@ package org.eclipse.lemminx.maven;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.maven.artifact.Artifact;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.model.Repository;
 import org.apache.maven.plugin.InvalidPluginDescriptorException;
@@ -94,14 +96,21 @@ public class MavenPluginUtils {
 			pluginKey += artifactId.get();
 		}
 		Plugin plugin = project.getPlugin(pluginKey);
-		if (plugin == null) {
+		if (plugin == null && project.getPluginManagement() != null) {
 			plugin = project.getPluginManagement().getPluginsAsMap().get(pluginKey);
+			
+			if (plugin == null && artifactId.isPresent()) {
+				//pluginArtifactMap will be empty if PluginManagement is null
+				for (Entry <String, Artifact> entry : project.getPluginArtifactMap().entrySet() ) {
+					if (entry.getValue().getArtifactId().equals(artifactId.get())) {
+						plugin = project.getPlugin(entry.getKey());
+					}
+				}
+			}
 		}
+		
 		if (plugin == null) {
-			throw new InvalidPluginDescriptorException("Incomplete GAV", Collections.emptyList());
-		}
-		if (plugin.getVersion() == null) {
-			throw new InvalidPluginDescriptorException("Missing version for " + plugin.toString(), Collections.emptyList());
+			throw new InvalidPluginDescriptorException("Unable to resolve " + pluginKey,  Collections.emptyList());
 		}
 
 		return pluginManager.getPluginDescriptor(plugin, project.getPluginRepositories().stream()
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
@@ -35,44 +35,44 @@ public class PluginValidator {
 		this.repoSession = repoSession;
 		this.pluginManager = pluginManager;
 	}
-	
-	public boolean validatePluginResolution(DiagnosticRequest diagnosticRequest) {
+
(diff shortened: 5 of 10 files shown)


lemminx-maven/pom.xml
```diff
@@ -115,7 +115,7 @@
 		<groupId>org.apache.maven.plugins</groupId>
 		<artifactId>maven-shade-plugin</artifactId>
 		<version>3.2.2</version>
-		<type>pom</type>
+		<type>test-jar</type>
```
AObuchow, line 118: Is this necessary? Double-check if `jar` will suffice.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DiagnosticRequest.java
```diff
@@ -62,6 +83,16 @@ public Range getRange() {
 				((DOMElement) node).getEndTagOpenOffset(), xmlDocument);
 	}
 
+	public void addDiagnostic(String errorMessage, DiagnosticSeverity severity) {
+		Diagnostic diagnostic = new Diagnostic(this.getRange(), errorMessage, severity,
+				this.getXMLDocument().getDocumentURI(), "XML");
+
+		// TODO: Should diagnostics just be a set?
```
AObuchow, line 90: Do this `TODO`?

mickaelistria, line 90: Yes, probably. Especially since they're actually not ordered.

AObuchow: Sounds good, will do.

AObuchow:
Maybe also assert that the range of the diagnostic for unresolvable plugins is correct (in the tests).

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -82,80 +76,66 @@ public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
...
 					| InvalidPluginDescriptorException e) {
 				e.printStackTrace();
 				// A diagnostic was already added in validatePluginResolution()
 			}
 		}
-		return null;
 	}
 
-	private static Diagnostic internalValidateGoal(DiagnosticRequest diagnosticReq, PluginDescriptor pluginDescriptor) {
+	private static void internalValidateGoal(DiagnosticRequest diagnosticReq, PluginDescriptor pluginDescriptor) {
```
mickaelistria, line 112: Functions (which don't have side effects) usually leads to easier maintenance than methods like this one that are built on side effects.
So I'd suggest the method becomes something like `private Optional<Diagnostic> validateGoal(DOMNode requestNode, descriptor)`, so consumers could just invoke `validateGoal(node, descriptor).ifPresent(request::addDiagnostic)`. This would also allow easier fine grain testing if necessary.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -82,80 +76,66 @@ public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
...
-
-		return new Diagnostic(diagnosticReq.getRange(), "Invalid goal for this plugin: " + goal,
-				DiagnosticSeverity.Warning, diagnosticReq.getXMLDocument().getDocumentURI(), "XML");
-
+		diagnosticReq.addDiagnostic("Invalid goal for this plugin: " + goal, DiagnosticSeverity.Warning);
 	}
 
-	private static Diagnostic internalValidateConfiguration(DiagnosticRequest diagnosticReq,
-			List<Parameter> parameters) {
+	private static void internalValidateConfiguration(DiagnosticRequest diagnosticReq, List<Parameter> parameters) {
```
mickaelistria, line 127: same as above

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -82,80 +76,66 @@ public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
...
-				}
+				DiagnosticRequest childDiagnosticReq = new DiagnosticRequest(childNode, diagnosticRequest);
+				internalValidateConfiguration(childDiagnosticReq, parameters);
 			}
 		}
-		return null;
 	}
 
-	public Diagnostic validateGoal(DiagnosticRequest diagnosticRequest) {
+	public void validateGoal(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 89: Functions (which don't have side effects) usually leads to easier maintenance than methods like this one that are built on side effects.
So I'd suggest the method becomes something like private Optional<Diagnostic> validateGoal(DOMNode requestNode, descriptor), so consumers could just invoke validateGoal(node, descriptor).ifPresent(request::addDiagnostic). This would also allow easier fine grain testing if necessary.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -35,44 +34,39 @@ public PluginValidator(MavenProjectCache cache, RepositorySystemSession repoSess
...
+								+ MavenPluginUtils.LINE_BREAK + "Additional information: " + errorMessage,
+						DiagnosticSeverity.Warning);
 			});
 			return false;
 		}
 		return true;
 	}
 
-	public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
+	public void validateConfiguration(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 61: See below, about returning values instead of using side effects on parameters.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDiagnosticParticipant.java
```diff
@@ -50,29 +51,17 @@ public void doDiagnostics(DOMDocument xmlDocument, List<Diagnostic> diagnostics,
 		
 		projectCache.getProblemsFor(xmlDocument).stream().map(this::toDiagnostic).forEach(diagnostics::add);
 		DOMElement documentElement = xmlDocument.getDocumentElement();
-		HashMap<String, Function<DiagnosticRequest, Diagnostic>> tagDiagnostics = configureDiagnosticFunctions(
-				xmlDocument);
+		HashMap<String, Consumer<DiagnosticRequest>> tagDiagnostics = configureDiagnosticFunctions();
```
mickaelistria, line 54: See comments below, I think Function drive to better code than consumers.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
AObuchow: +1 I knew something felt wrong about how the diagnostics were being added.

Will make the fix + other related fixes.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -94,14 +96,21 @@ public static PluginDescriptor getContainingPluginDescriptor(IPositionRequest re
...
 		Plugin plugin = project.getPlugin(pluginKey);
-		if (plugin == null) {
+		if (plugin == null && project.getPluginManagement() != null) {
 			plugin = project.getPluginManagement().getPluginsAsMap().get(pluginKey);
+			
+			if (plugin == null && artifactId.isPresent()) {
+				//pluginArtifactMap will be empty if PluginManagement is null
+				for (Entry <String, Artifact> entry : project.getPluginArtifactMap().entrySet() ) {
+					if (entry.getValue().getArtifactId().equals(artifactId.get())) {
+						plugin = project.getPlugin(entry.getKey());
```
AObuchow, line 106: This is to fix #74

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -136,16 +137,28 @@ public void testPluginDiagnosticMissingPlugin()
 		DOMDocument document = createDOMDocument("/pom-plugin-diagnostic-missing-plugin.xml", languageService);
 		System.out.println(languageService.doDiagnostics(document, () -> {
```
AObuchow, line 138: remove system.out

AObuchow:
@mickaelistria this should be good to review now. Sorry for the kind of large change (addressing two bugs at once).

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/LocalPluginTest.java
```diff
@@ -136,16 +137,29 @@ public void testPluginDiagnosticMissingPlugin()
...
-				"Incomplete GAV")));
+		}, new XMLValidationSettings()).size() == 1); 
+	}
+	
+	@Test
+	public void testDiagnosticMissingGroupId() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-plugin-diagnostic-missing-groupid.xml", languageService);
+		List<Diagnostic> diagnostics = languageService.doDiagnostics(document, () -> {
+		}, new XMLValidationSettings());
+		System.out.println(diagnostics);
```
AObuchow, line 153: Remove before merge.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -82,80 +79,73 @@ public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
 			// A diagnostic was already added in validatePluginResolution()
 		}
 		if (parameters.isEmpty()) {
-			return null;
+			return Optional.ofNullable(null);
```
mickaelistria, line 82: Wouldn't `Optional.empty()` be better here?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -82,80 +79,73 @@ public Diagnostic validateConfiguration(DiagnosticRequest diagnosticRequest) {
...
+		
+		return Optional.of(diagnostics);
 	}
 
-	public Diagnostic validateGoal(DiagnosticRequest diagnosticRequest) {
+	public Optional<List<Diagnostic>> validateGoal(DiagnosticRequest diagnosticRequest) {
 		DOMNode node = diagnosticRequest.getNode();
 		if (node == null) {
-			return null;
+			return Optional.ofNullable(null);
```
mickaelistria, line 97: `Optional.empty()` ?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/SubModuleValidator.java
```diff
@@ -34,21 +32,15 @@ public void setPomFile(File pomFile) throws FileNotFoundException, IOException,
 
 	public Diagnostic validateSubModuleExistence(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 33: This could also be an `Optional<Diagnostic>`, or even an `Optional<List<Diagnostic>>` so this could be referenced in MavenDiagnosticParticipant.tagDiagnostics.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
```diff
@@ -35,44 +35,41 @@ public PluginValidator(MavenProjectCache cache, RepositorySystemSession repoSess
 		this.repoSession = repoSession;
 		this.pluginManager = pluginManager;
 	}
-	
-	public boolean validatePluginResolution(DiagnosticRequest diagnosticRequest) {
+
+	public Diagnostic validatePluginResolution(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 39: This could also be an `Optional<Diagnostic>`, or even an `Optional<List<Diagnostic>>` so this could be referenced in MavenDiagnosticParticipant.tagDiagnostics.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/VersionValidator.java
```diff
@@ -12,34 +12,24 @@
...
 import org.apache.maven.model.Dependency;
-import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
-import org.eclipse.lsp4j.Range;
 
 public class VersionValidator {
 
 	public static Diagnostic validateVersion(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 21: This could also be an `Optional<Diagnostic>`, or even an `Optional<List<Diagnostic>>` so this could be referenced in MavenDiagnosticParticipant.tagDiagnostics.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
AObuchow: +1

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/PluginValidator.java
AObuchow: Yes my bad :P

AObuchow:
@mickaelistria thanks, I addressed your comments.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/SubModuleValidator.java
```diff
@@ -32,23 +33,19 @@ public void setPomFile(File pomFile) throws FileNotFoundException, IOException,
...
 		String tagContent = null;
 		if (node.hasChildNodes()) {
 			tagContent = node.getChild(0).getNodeValue(); // tagContent is the module to validate eg.
 															// <module>tagContent</module>
 		}
 		if (node.hasChildNodes() && !model.getModules().contains(tagContent)) {
-			diagnostic = new Diagnostic(range, String.format("Module '%s' does not exist", tagContent),
-					DiagnosticSeverity.Error, xmlDocument.getDocumentURI(), "XML");
-			diagnosticRequest.getDiagnostics().add(diagnostic);
+			List<Diagnostic> diagnostics = new ArrayList<>();
```
mickaelistria, line 44: You can use `Collections.singletonList(diagnosticRequest.createDiagnostic(....))` instead of manually instantiating a new ArrayList. Singleton collections will consume slightly less memory.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/SubModuleValidator.java
```diff
@@ -32,23 +33,19 @@ public void setPomFile(File pomFile) throws FileNotFoundException, IOException,
 		model = mavenreader.read(new FileReader(pomFile));
 	}
 
-	public Diagnostic validateSubModuleExistence(DiagnosticRequest diagnosticRequest) {
+	public Optional<Object> validateSubModuleExistence(DiagnosticRequest diagnosticRequest) {
```
mickaelistria, line 36: Why not `Optional<List<Diagnostic>>` here?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/VersionValidator.java
```diff
@@ -8,38 +8,34 @@
...
 		Artifact artifact = null;
-		Range range = diagnosticRequest.getRange();
-		Diagnostic diagnostic = null;
 		try {
 			 artifact = new DefaultArtifact(model.getGroupId(), model.getArtifactId(), model.getVersion(), model.getScope(), model.getType(), model.getClassifier(), new DefaultArtifactHandler(model.getType()));
 			 if (!artifact.isSelectedVersionKnown()) {
-				diagnostic = new Diagnostic(range, "Version Error", DiagnosticSeverity.Error,
-						xmlDocument.getDocumentURI(), "XML");
-				diagnosticRequest.getDiagnostics().add(diagnostic);
+				 List<Diagnostic> diagnostics = new ArrayList<>();
```
mickaelistria, line 32: Same as above, you can probably use `Collections.singletonList()`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/SubModuleValidator.java
AObuchow: Thanks I've wondered what was the nicer way of doing that. Will adjust.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/SubModuleValidator.java
AObuchow: Oops, good catch.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/VersionValidator.java
AObuchow: > Some more minor comments.

Just addressed these.

> going functional here is a cool pattern and keeping a node-name/functon registry leads to nice code, we may consider this for other classes later.

Agreed, it's something to consider.

AObuchow:
>Some more minor comments.

Just addressed these.

> going functional here is a cool pattern and keeping a node-name/functon registry leads to nice code, we may consider this for other classes later.

Agreed, it's something to consider.

AObuchow:
Should be good now.

mickaelistria:
Looks good to merge IMO.

Issue #72, opened 2020-05-25
Setup Windows CI with GitHub Actions


AObuchow:
To help prevent future Windows-only regressions, it'd be nice to have a GitHub Actions Workflow setup to run the build + tests on Windows. In the past, LemMinX-Maven was already using GitHub Actions for CI, see [here](https://github.com/AObuchow/lsp4xml-extensions-maven/tree/master/.github/workflows).

PR #81, opened 2020-06-04
Fix Windows bug with paths test

### LINKED ISSUE #69: Unit tests failing on windows
Following is the test execution done in a azure pipeline with windows-latest vm.
[lemmix-windows-tests.txt](https://github.com/eclipse/lemminx-maven/files/4671457/lemmix-windows-tests.txt)

@AObuchow Shall we try to add a window github workflow as well ?

AObuchow:
Fix #69


AObuchow:
@gayanper this should fix the issue with the paths tests on Windows.

mickaelistria:
I actually don't think usage of `\` is desired in pom files usually. IMO, the tests are correct, it's more the actual completion should force usage of `/` as file separator.

AObuchow:
> I actually don't think usage of `\` is desired in pom files usually. IMO, the tests are correct, it's more the actual completion should force usage of `/` as file separator.

+1 agreed, will rework the fix

AObuchow:
@mickaelistria the new patch should hopefully be a more valid fix. I think it'd be a good idea for me to fix #72 before merging this fix, to ensure the fix is valid.

mickaelistria:
We now have a bit more time to do things slower and better. So if you prefer completing #72 first, feel free.

PR #93, opened 2020-06-19
Update to LemMinX 0.12.0

### LINKED ISSUE #92: Migrate to LemMinX 0.12.0
LemMinX 0.12.0 was released. LemMinX-Maven extension should adopt it as it has some improvements and API changes.

AObuchow:
StringUtils.findExpreBeforeAt was removed from LemMinX so it was moved
to SnippetRegistry.

Fix #92

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/snippets/SnippetRegistry.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/snippets/SnippetRegistry.java
@@ -119,7 +119,7 @@ public class SnippetRegistry {
 			return contextFilter.test(s.getContext());
 		}).map(snippet -> {
 			String filter = snippet.getPrefix();
-			int startOffset = StringUtils.findExprBeforeAt(document.getText(), filter, completionOffset);
+			int startOffset = findExprBeforeAt(document.getText(), filter, completionOffset);
 			if (startOffset == -1) {
 				startOffset = completionOffset;
 			}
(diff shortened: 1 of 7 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/snippets/SnippetRegistry.java
```diff
@@ -195,4 +195,30 @@ private static Range getReplaceRange(int replaceStart, int replaceEnd, TextDocum
 		}
 		return new Range(document.positionAt(replaceStart), document.positionAt(replaceEnd));
 	}
+	
+	private static int findExprBeforeAt(String text, String expr, int offset) {
```
AObuchow, line 199: @angelozerr StringUtilsfindExpreBeforeAt() was removed upstream in https://github.com/eclipse/lemminx/commit/811cb64feb30d179b023d08ea45376df75ad0845#diff-d38ece0211b24629cca68b0dc5cbc294, do you have any suggested upstream alternative? Otherwise we'll just keep it downstream.

mickaelistria: I don't think we use Snippet, nor that we plan to use them, so if it's the main issue, just get rid of the SnippetRegistry.

AObuchow: I don't think it's really an issue (whether we keep it or remove). Right now it seems like the build is failing due to the fact LemMinX 0.12.0 hasn't been released on the Eclipse repos. 

However, I can remove it if we don't plan to use it.

AObuchow: Actually.. one of the builds was successful so once I rerun them I can remove snippet registry (or keep it) and this will be good for an eventual merge.

angelozerr: The idea is that SnippetRegistry is now managed inside LemMinx, so please  remove this class.

The main idea is to contribute to snippet registry with SPI.

I suggest that you see xml-model snippet PR https://github.com/eclipse/lemminx/pull/752

Don't hesitate to ping me if you need some help.

angelozerr: > I don't think we use Snippet, nor that we plan to use them, so if it's the main issue, just get rid of the SnippetRegistry.

You are using snippets (it was in my original POC). See https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/resources/org/eclipse/lemminx/maven/pom-snippets.json

mickaelistria:
@AObuchow since the "pr-head" passes, I think this can be merged right now.

AObuchow:
> @AObuchow since the "pr-head" passes, I think this can be merged right now.

Sure, merging now.

PR #99, opened 2020-07-02
Allow environment variable maven properties to be parsed

### LINKED ISSUE #94: Environment variables aren't being parsed correctly
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=564500 for more information. (Will add more details to this bug shortly)

AObuchow:
Fix #94

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
@@ -19,15 +19,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
-import org.apache.maven.artifact.InvalidRepositoryException;
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
 import org.apache.maven.artifact.repository.MavenArtifactRepository;
 import org.apache.maven.artifact.repository.layout.DefaultRepositoryLayout;
 import org.apache.maven.execution.MavenExecutionRequest;
-import org.apache.maven.execution.MavenExecutionRequestPopulationException;
 import org.apache.maven.model.Build;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.building.DefaultModelProblem;
@@ -44,11 +43,10 @@ import org.apache.maven.project.ProjectBuilder;
 import org.apache.maven.project.ProjectBuildingException;
 import org.apache.maven.project.ProjectBuildingRequest;
 import org.apache.maven.project.ProjectBuildingResult;
-import org.apache.maven.settings.io.SettingsParseException;
+import org.apache.maven.properties.internal.EnvironmentUtils;
 import org.codehaus.plexus.PlexusContainer;
 import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
-import org.eclipse.aether.DefaultRepositorySystemSession;
 import org.eclipse.lemminx.dom.DOMDocument;
 
 public class MavenProjectCache {
@@ -60,10 +58,10 @@ public class MavenProjectCache {
 
 	private final MavenExecutionRequest mavenRequest;
 	MavenXpp3Reader mavenReader = new MavenXpp3Reader();
-	private DefaultRepositorySystemSession repositorySystemSession;
 	private ProjectBuilder projectBuilder;
 
 	private final List<Consumer<MavenProject>> projectParsedListeners = new ArrayList<>();
+	private Properties systemProperties;
 
 	public MavenProjectCache(PlexusContainer container, MavenExecutionRequest mavenRequest) {
 		this.mavenRequest = mavenRequest;
@@ -71,6 +69,10 @@ public class MavenProjectCache {
 		this.lastCheckedVersion = new HashMap<URI, Integer>();
 		this.projectCache = new HashMap<URI, MavenProject>();
 		this.problemCache = new HashMap<URI, Collection<ModelProblem>>();
+		systemProperties = new Properties();
+		EnvironmentUtils.addEnvVars(systemProperties);
+		systemProperties.putAll(System.getProperties());
+		initializeMavenBuildState();
 	}
 
 	/**
@@ -110,18 +112,10 @@ public class MavenProjectCache {
 		if (lastKnownVersionMavenProject != null) {
 			return Optional.of(lastKnownVersionMavenProject);
 		}
-		if (mavenRequest == null) {
-			try {
-				initializeMavenBuildState();
-			} catch (Exception e) {
-				// TODO Auto-generated catch block
-				e.printStackTrace();
-			}
-		}
 		ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-		request.setSystemProperties(System.getProperties());
+		request.setSystemProperties(systemProperties);
 		request.setLocalRepository(mavenRequest.getLocalRepository());
-		request.setRepositorySession(getRepositorySystemSession());
+		request.setRepositorySession(MavenPlugin.getRepositorySystemSession());
 		request.setResolveDependencies(true);
 		try {
 			MavenProject project = projectBuilder.build(file, request).getProject();
@@ -140,13 +134,10 @@ public class MavenProjectCache {
 		URI uri = URI.create(document.getDocumentURI());
 		Collection<ModelProblem> problems = new ArrayList<ModelProblem>();
 		try {
-			if (mavenRequest == null) {
-				initializeMavenBuildState();
-			}
 			ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-			request.setSystemProperties(System.getProperties());
+			request.setSystemProperties(systemProperties);
 			request.setLocalRepository(mavenRequest.getLocalRepository());
-			request.setRepositorySession(getRepositorySystemSession());
+			request.setRepositorySession(MavenPlugin.getRepositorySystemSession());
 			request.setResolveDependencies(true);
 			ProjectBuildingResult buildResult = projectBuilder.build(new FileModelSource(new File(uri)) {
 				@Override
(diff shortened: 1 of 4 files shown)


lemminx-maven/pom.xml
```diff
@@ -89,6 +89,16 @@
 			<artifactId>wagon-http</artifactId>
 			<version>3.3.4</version>
 		</dependency>
+	<dependency>
```
AObuchow, line 92: @mickaelistria these aren't required for the fix, however I found them in the m2e runtime. I can remove them realistically if they're not providing any improvements.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -119,9 +122,18 @@ private void check(DOMDocument document) {
 			}
 		}
 		ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-		request.setSystemProperties(System.getProperties());
+		Properties systemProperties = new Properties();
+		EnvironmentUtils.addEnvVars(systemProperties);
```
AObuchow, line 126: This is the actual fix.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -230,12 +273,14 @@ public void addProjectParsedListener(Consumer<MavenProject> listener) {
...
 	}
 
+	@Deprecated
 	public DefaultRepositorySystemSession getRepositorySystemSession() {
 		try {
 			initializeMavenBuildState();
 		} catch (Exception e) {
 			e.printStackTrace();
 		}
+		// TODO: How does this initialize the repository system session?
```
AObuchow, line 283: @mickaelistria am I missing something here? Or is it ok to remove/deprecate this method? I'm now passing the repositorySystemSession from MavenPlugin

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -115,6 +116,17 @@ public void testMissingArtifactIdError()
...
 		assertEquals(Collections.emptyList(), languageService.doDiagnostics(document, () -> {}, new XMLValidationSettings()));
 	}
+	
+	@Test
+	public void testSystemPathDiagnosticBug()
+			throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-environment-variable-property.xml", languageService);
+		List<Diagnostic> diagnostics = languageService.doDiagnostics(document, () -> {
+		}, new XMLValidationSettings());
+		System.out.println(System.getenv("JAVA_HOME"));
```
AObuchow, line 126: Remove

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -115,6 +116,17 @@ public void testMissingArtifactIdError()
...
+		System.out.println(diagnostics);
```
AObuchow, line 127: Remove

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -119,9 +122,18 @@ private void check(DOMDocument document) {
 			}
 		}
 		ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-		request.setSystemProperties(System.getProperties());
+		Properties systemProperties = new Properties();
+		EnvironmentUtils.addEnvVars(systemProperties);
+		copyProperties(systemProperties, System.getProperties());
```
mickaelistria, line 127: `systemProperties.putAll(System.getProperties())`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: Yes, it seems good to remove ti.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -144,9 +156,18 @@ private void parseAndCache(DOMDocument document) {
 				initializeMavenBuildState();
 			}
 			ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-			request.setSystemProperties(System.getProperties());
+			Properties systemProperties = new Properties();
+			EnvironmentUtils.addEnvVars(systemProperties);
+			copyProperties(systemProperties, System.getProperties());
```
mickaelistria, line 161: `systemProperties.putAll(System.getProperties())`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: Actually, you may want to remove initializeMavenBuildState by the way and make the projectBuilder and the DISABLE_GLOBAL_MODEL_CACHE_SYSTEM_PROPERTY system property set in the constructor.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: Couldn't/shouldn't this properties be initialized only once in constructor rather than on every request? I don't think either the system.getProperties() nor the environment can nor should change during execution of the LS (in worst case, people would just have to restart it), so it may be good to store it once and for all.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: Good call, I'll try that

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: +1

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: +1, done

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: I ended up moving initializeMavenBuildState to the constructor.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -213,13 +204,17 @@ public InputStream getInputStream() throws IOException {
 		problemCache.put(uri, problems);
 	}
 
-	private void initializeMavenBuildState() throws ComponentLookupException, InvalidRepositoryException,
-			SettingsParseException, IOException, MavenExecutionRequestPopulationException {
-		if (projectBuilder != null) {
+	private void initializeMavenBuildState(MavenExecutionRequest mavenRequest) {
+		if (projectBuilder != null || mavenRequest == null) {
```
AObuchow, line 208: should be `mavenRequest != null`

AObuchow: @mickaelistria actually.. I'm a bit confused by this. It seems the tests are passing and tests don't pass if I change the if statement to  `if (projectBuilder != null || mavenRequest != null) {` (the tests are passing with how it is currently).

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -213,13 +204,17 @@ public InputStream getInputStream() throws IOException {
 		problemCache.put(uri, problems);
 	}
 
-	private void initializeMavenBuildState() throws ComponentLookupException, InvalidRepositoryException,
-			SettingsParseException, IOException, MavenExecutionRequestPopulationException {
```
AObuchow, line 217: Also I took out these "throws" as it seems only `ComponentLookupException` can be thrown?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: Why are you adding `mavenRequest != null`? As far as I understand, the request can totally be non-null while the projectBuilder can still be null.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: OK

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
```diff
@@ -110,18 +112,10 @@ private void check(DOMDocument document) {
 		if (lastKnownVersionMavenProject != null) {
 			return Optional.of(lastKnownVersionMavenProject);
 		}
-		if (mavenRequest == null) {
```
AObuchow, line 113: @mickaelistria this is the line that made me assume the maven build state shouldn't be initialized if the mavenRequest is non-null. 

I assume the way I have it currently is ok? Ie. don't initialize the Maven Build State if the mavenRequest is null.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: See my other comment. It's the pre-existing code that's confusing me a bit

AObuchow: I think the bigger question is: should `initializeMavenbuildState` depend on whether the mavenRequest is null or not?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: I think we should care about the mavenRequest at this point.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
mickaelistria: I don't think so, let's try without the condition.

AObuchow: Tests seem to pass. Going to merge this.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenProjectCache.java
AObuchow: So do you think the maven build state should still be initialized in all areas as previously? Or only in the constructor is OK? (as I have it currently)

mickaelistria:
we need to make sure it's initialized only once, so the constructor seems to be a good location.

Issue #97, opened 2020-07-01
Maven property hover doesn't always work


AObuchow:
Using Eclipse IDE 4.16 on the following pom, I'm unable to get maven property hover working.

```XML
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<modelVersion>4.0.0</modelVersion>

	<groupId>org.test</groupId>
	<artifactId>test</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<packaging>pom</packaging>

	<properties>
		<myProperty>$</myProperty>
	</properties>
	<dependencies>
		<dependency>
			<groupId>org.${myProperty}.${project.version}</groupId>
		</dependency>
	</dependencies>
</project>
```

 I am however able to get ctrl + click to open the property declaration, and I can see that the property is getting resolved by the maven project builder (as a diagnostic is created stating `'dependencies.dependency.groupId' for org.$.0.0.1-SNAPSHOT:null:jar with value 'org.$.0.0.1-SNAPSHOT' does not match a valid id pattern.`).

AObuchow:
Running `SimpleModelTest.testPropertyHover`, I can see that the kind "plaintext" is used when the type is MarkupContent:
```
Hover [
  contents = Either [
    left = null
    right = MarkupContent [
    kind = "plaintext" <----------------------
    value = "Property: myProperty\n\nValue: $\n\n"
  ]
  ]
  range = Range [
    start = Position [
      line = 15
      character = 12
    ]
    end = Position [
      line = 15
      character = 48
    ]
  ]
]
```

Maybe this could be causing the bug @mickaelistria ?

mickaelistria:
I don't think the "kind" is supposed to cause a bug, and if it does, it's probably more a bug in LSP4E. Did you try debugging m2e and LSP4E to verify whether the "kind" cause the bug?

AObuchow:
@mickaelistria I agree that if this causes an issue, it should be a LSP4E bug.

I haven't tried debugging m2e + LSP4E to verify this yet. I was thinking of just building the lemminx-maven jar locally and testing it in m2e to see if changing the "kind" fixes the issue.

[...]

PR #100, opened 2020-07-02
Maven property hovers should MarkupContent use "markdown" kind

AObuchow:
Part of #97

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
@@ -327,7 +327,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 			for (Entry<String, String> prop : allProps.entrySet()) {
 				String mavenProperty = prop.getKey();
 				if (property.equals(mavenProperty)) {
-					return new Hover(new MarkupContent("plaintext", "Property: " + mavenProperty
+					return new Hover(new MarkupContent("markdown", "Property: " + mavenProperty
 							+ MavenPluginUtils.LINE_BREAK + "Value: " + prop.getValue() + MavenPluginUtils.LINE_BREAK));
 				}
 			}


mickaelistria:
I don't think it's correct. The Maven parameter documentation is not markdown, so it shouldn't be marked as being markdown.
It could be a bug in LSP4E not properly rendering the documentation in case it's not markdown; but I don't think the LS should be where we try to hack things.

AObuchow:
@mickaelistria the `MavenPluginUtils.LINE_BREAK` which is used in the maven property hover is `\n\n` which is markdown. I do think that an investigation in LSP4E needs to be made, however, this patch seems correct to me.

mickaelistria:
I think using markdown is strictly incorrect as Maven doc seems to say that documentations are typically using Javadoc format. However, as Javadoc is partly HTML and Markdown renderer is supposed to partly render HTML, we can make the switch.
ie it's incorrect, but less incorrect than plaintext.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -327,7 +327,7 @@ private Hover collectProperty(IPositionRequest request, String property) {
 			for (Entry<String, String> prop : allProps.entrySet()) {
 				String mavenProperty = prop.getKey();
 				if (property.equals(mavenProperty)) {
-					return new Hover(new MarkupContent("plaintext", "Property: " + mavenProperty
+					return new Hover(new MarkupContent("markdown", "Property: " + mavenProperty
```
angelozerr, line 330: You should take care if client can support markdown.

To do that:

`boolean markdown = request.canSupportMarkupKind(MarkupKind.MARKDOWN);`

After that I suggest you that you do like we have done for entites doc (for hover and completion). See https://github.com/eclipse/lemminx/blob/374ea32776f24a910f6a2541d0f66dd21cff1266/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/entities/EntitiesDocumentationUtils.java#L91

mickaelistria:
ah sorry, I just realized that your patch is about _properties_ while I had in mind _parameters_. So yes, as the properties don't have a specific documentation, it's fine to use markdown.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
mickaelistria: At the moment, I think we can just not care too much about such clients as they should be marginal. We'll work on that when we have a request for it from some integrator and we consider it's high priority. Realistically, it's more the kind of things we will let the people who want it implement.

angelozerr:
> I think using markdown is strictly incorrect as Maven doc seems to say that documentations are typically using Javadoc format. However, as Javadoc is partly HTML and Markdown renderer is supposed to partly render HTML, we can make the switch.
ie it's incorrect, but less incorrect than plaintext.

We had the same problem with XSD documentation. To fix that we generate HTML content and convert the HTML to markdown.

I suggest you that you generate your doc in HTML and after that you consume https://github.com/eclipse/lemminx/blob/374ea32776f24a910f6a2541d0f66dd21cff1266/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/MarkupContentFactory.java#L41

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
angelozerr: If you use https://github.com/eclipse/lemminx/blob/374ea32776f24a910f6a2541d0f66dd21cff1266/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/MarkupContentFactory.java#L41 everything should be supported

AObuchow:
Thank you for the advice @angelozerr I like that approach. As we are creating markdown in a few areas of the code, I feel that a separate bug should be made to use this approach everywhere in LemMinX-Maven (not just in this commit). I'll go write that now :)

PR #103, opened 2020-07-08
Ensure remote artifacts download on artifactId hover from central

### LINKED ISSUE #57: Download artifacts from remote repositories on-demand (hover, completion, validation...)`
When trying to validate or get assistance on some plugin or dependency that's not available in local repo and that's necessary to have locally for smart results, it should be downloaded locally so the LS would then provide smartest results.

AObuchow:
Fix #57


lemminx-maven/src/test/resources/pom-remote-artifact-download-hover.xml
```diff
@@ -0,0 +1,28 @@
...
+
+	<dependencies>
+	<dependency>
+		<groupId>org.glassfish.jersey</groupId>
+		<artifactId>project</artifactId>
+		<version>2.19</version>
+	</dependency>
+	</dependencies>
+
+	<repositories>
```
AObuchow, line 20: Try removing the <repository> to see if this test still passes? ie. is lemminx-maven actually downloading the artifact from maven central

AObuchow: @mickaelistria this confirms that the downloads are occurring from maven central and not other repositories listed in the pom.

Issue #110, opened 2020-07-10
Duplicate plugin configuration completion items


AObuchow:
In the following pom:
```XML
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<modelVersion>4.0.0</modelVersion>
	<groupId>demo</groupId>
	<artifactId>demo</artifactId>
	<version>0.0.1-SNAPSHOT</version>
  
	<build>
		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.8.1</version>
				<configuration>
				<comp>
				</configuration>
			</plugin>
		</plugins>
	</build>
</project>
```

I'm getting duplicate plugin configuration proposals. They seem to be exactly the same.

![image](https://user-images.githubusercontent.com/10300119/87172422-37825400-c2a2-11ea-8d57-09f82471e6c6.png)
![image](https://user-images.githubusercontent.com/10300119/87172507-58e34000-c2a2-11ea-9a41-5d83a79d60d1.png)

AObuchow:
Note that this issue occurred once before https://github.com/eclipse/lemminx-maven/issues/53, the fix was https://github.com/eclipse/lemminx-maven/commit/e80b44bd476dffacacffd75e500932750c53062e

AObuchow:
@angelozerr @fbricon  I believe an upstream fix in LemMinX might be best. From what I recall the issue is that the same completion is being added in [`onTagOpen`](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java#L125) and [`onXMLContent`](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java#L268).

There's no way to retrieve the completion items already added in the response. However, rather than letting clients check to see if they're adding a duplicate, maybe the upstream/internal container for completion responses should be a set? This would prevent other XML extensions from encountering the same bug in the future.

AObuchow:
[...]

PR #111, opened 2020-07-10
Test for duplicate plugin configuration parameters in completion

AObuchow:
Part of #110

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
@@ -265,13 +265,6 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 		case "goal":
 			collectGoals(request).forEach(response::addCompletionItem);
 			break;
-		case "configuration":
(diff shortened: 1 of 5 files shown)


AObuchow:
Test is intended to fail currently.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
```diff
@@ -265,13 +265,6 @@ public void onXMLContent(ICompletionRequest request, ICompletionResponse respons
 		case "goal":
 			collectGoals(request).forEach(response::addCompletionItem);
 			break;
-		case "configuration":
```
AObuchow, line 268: @mickaelistria it seems this wasn't necessary, tests are passing without it and things seemed to work when user testing. Any worries about removing this?

mickaelistria:
So do I get it right that when completing

```xml
  <configuration>
     |(completeHere)
  </configuration>
```
LemMinX invokes both onXMLContent and onTagOpen to gather the results?
If so, it's a good thing LemMinX does that, and feel free to remove this piece of code if it causes bugs. Maybe we can also identify other cases that belong to this pattern then, and move them from onXMLCotent to onTagOpen, it seems like anything that inserts a tag could be a good candidate.

AObuchow:
@mickaelistria that seems correct to me, based on my understanding. @angelozerr could you maybe confirm?

PR #113, opened 2020-07-10
Env var property hover

AObuchow:
Fix #106

PR #115, opened 2020-07-15
Environment variable property hover + completion

AObuchow:
Fix #113

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
@@ -513,6 +513,10 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 				CompletionItem item = toTextCompletionItem(request, "${" + property.getKey() + '}');
 				item.setDocumentation("Default Value: " + (property.getValue() != null ? property.getValue() : "unknown"));
 				item.setKind(CompletionItemKind.Property);
+				if (property.getKey().contains("env.")) {
+					// We don't want environment variables at the top of the completion proposals
+					item.setSortText("${zzz" + property.getKey() + "}");
+				}
 				return item;
 			} catch (BadLocationException e) {
 				e.printStackTrace();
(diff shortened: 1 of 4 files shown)


AObuchow:
both commits should be squashed into one

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -62,6 +64,7 @@
 	private final RepositorySystemSession repoSession;
 	private final MavenSession mavenSession;
 	private final BuildPluginManager buildPluginManager;
+	private static Properties environmentProperties;
```
AObuchow, line 67: I'm not so sure about making this static

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
```diff
@@ -513,6 +513,10 @@ private CompletionItem toTag(String name, MarkupContent description, ICompletion
 				CompletionItem item = toTextCompletionItem(request, "${" + property.getKey() + '}');
 				item.setDocumentation("Default Value: " + (property.getValue() != null ? property.getValue() : "unknown"));
 				item.setKind(CompletionItemKind.Property);
+				if (property.getKey().contains("env.")) {
+					// We don't want environment variables at the top of the completion proposals
+					item.setSortText("${z}");
```
AObuchow, line 518: This is a bit of a hack but it makes the completions rank last.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
mickaelistria: I think it's OK.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
mickaelistria: Maybe something like `"${zzz" + property.getKey() + "}"` would be better as it would reduce chances of a property starting with z being shown at wrong position, and would still allow the envVars to be sorted.

AObuchow: +1

PR #116, opened 2020-07-15
Fix incorrect range on maven property definition link

### LINKED ISSUE #107: Definition range for maven properties is incorrect
Doing a ctrl+click on a maven property will create a link for the entire node's text when in reality only the maven property should get a link.
image

AObuchow:
Fix #107

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
@@ -18,6 +18,7 @@ import java.util.stream.Collectors;
 import org.apache.maven.Maven;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.project.MavenProject;
+import org.eclipse.lemminx.commons.BadLocationException;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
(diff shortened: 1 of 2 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenDefinitionParticipant.java
```diff
@@ -124,7 +126,22 @@ private LocationLink findMavenPropertyLocation(IDefinitionRequest request) {
 		if (propertyDeclaration == null) {
 			return null;
 		}
-		return toLocation(childProj.getFile(), propertyDeclaration, request.getNode());		
+		
+
+		String mavenPropertyText = "${" + mavenProperty + "}";
+		Position propertyStart;
+		try {
+			propertyStart = new Position(request.getPosition().getLine(),
```
AObuchow, line 134: This is quite messy right now..

AObuchow:
This still doesn't cover the case where two maven properties are use in the same tag's text. The second maven property won't have any definition link.

mickaelistria:
I don't fully get the case you're mentioning. Can you please give an example?
In any case,as long as the result is better than current state, feel free to merge, even if there are some possible glitches (as long as those glitches are not regression).

AObuchow:
@mickaelistria the case is trying to ctrl+click on "bar" in the following pom:
```XML
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

	<groupId>org.test</groupId>
	<artifactId>test</artifactId>
	<packaging>pom</packaging>

	<properties>
		<foo>$</foo>
		<bar>test</bar>
	</properties>
	<contributors>
		<contributor>
			<name>${foo}   ${bar}</name>
		</contributor>
	</contributors>
</project>
```

I'll improve this patch a bit, see if I can fix the described case and if not, merge and make a new issue for it.

PR #119, opened 2020-07-22
Test hover when multiple properties are in a tag

### LINKED ISSUE #117: Hover with multiple properties in the same tag doesn't work
Hovering on "${bar}" in the following pom gives a hover for "foo":
```XML
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

	<groupId>org.test</groupId>
	<artifactId>test</artifactId>
	<packaging>pom</packaging>

	<properties>
		<foo>$</foo>
		<bar>test</bar>
	</properties>
	<contributors>
		<contributor>
			<name>${foo}   ${bar}</name>
		</contributor>
	</contributors>
</project>
```

image

AObuchow:
Fix #117

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
@@ -28,6 +28,8 @@ import java.util.concurrent.TimeoutException;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.execution.MavenSession;
@@ -54,9 +56,11 @@ import org.eclipse.lemminx.maven.searcher.RemoteRepositoryIndexSearcher;
 import org.eclipse.lemminx.services.extensions.IHoverParticipant;
 import org.eclipse.lemminx.services.extensions.IHoverRequest;
 import org.eclipse.lemminx.services.extensions.IPositionRequest;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
 import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.MarkupContent;
 import org.eclipse.lsp4j.MarkupKind;
+import org.eclipse.lsp4j.Range;
 
 public class MavenHoverParticipant implements IHoverParticipant {
 	private final MavenProjectCache cache;
@@ -299,7 +303,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 				|| (grandParent != null && "parent".equals(grandParent.getLocalName()));
 		
 		
-		String mavenProperty = getMavenPropertyInRequest(request);
+		Pair<Range, String> mavenProperty = getMavenPropertyInRequest(request);
 		if (mavenProperty != null) {
 			return collectProperty(request, mavenProperty);
 		}
@@ -321,7 +325,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 	}
 
 
-	public static String getMavenPropertyInRequest(IPositionRequest request) {
+	public static Pair<Range, String> getMavenPropertyInRequest(IPositionRequest request) {
 		DOMNode tag = request.getNode();
 		String tagText = tag.getNodeValue();
 		if (tagText == null) {
@@ -339,15 +343,19 @@ public class MavenHoverParticipant implements IHoverParticipant {
 		int indexCloseBefore = beforeHoverText.lastIndexOf('}');
 		int indexCloseAfter = afterHoverText.indexOf('}');
 		if (indexOpen > indexCloseBefore) {
-			return tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+
+			String propertyText = tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+			int textStart = request.getNode().getStart();
+			Range propertyRange = XMLPositionUtility.createRange(textStart + indexOpen + 2,
+					textStart + indexCloseAfter - 1, request.getXMLDocument());
+			return new ImmutablePair<Range, String>(propertyRange, propertyText);
 		}
 		return null;
 	}
 
-	private Hover collectProperty(IHoverRequest request, String property) {
+	private Hover collectProperty(IHoverRequest request, Pair<Range, String> property) {
 		boolean supportsMarkdown = request.canSupportMarkupKind(MarkupKind.MARKDOWN);
 		String lineBreak = MarkdownUtils.getLineBreak(supportsMarkdown);
-		
 		DOMDocument doc = request.getXMLDocument();
 		MavenProject project = cache.getLastSuccessfulMavenProject(doc);
 		if (project != null) {
@@ -356,13 +364,15 @@ public class MavenHoverParticipant implements IHoverParticipant {
 
 			for (Entry<String, String> prop : allProps.entrySet()) {
 				String mavenProperty = prop.getKey();
-				if (property.equals(mavenProperty)) {
+				if (property.getRight().equals(mavenProperty)) {
 					String message = toBold.apply("Property: ") + mavenProperty 
 							+ lineBreak
 							+ toBold.apply("Value: ") + prop.getValue() 
 							+ lineBreak;
 					
-					return new Hover(new MarkupContent(supportsMarkdown ? MarkupKind.MARKDOWN : MarkupKind.PLAINTEXT, message));
+					Hover hover = new Hover(new MarkupContent(supportsMarkdown ? MarkupKind.MARKDOWN : MarkupKind.PLAINTEXT, message));
+					hover.setRange(property.getLeft());
+					return hover;
 				}
 			}
 		}
--- a/lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
+++ b/lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
@@ -192,6 +192,48 @@ public class SimpleModelTest {
 	}
 	
 	@Test
+ 	public void testMultiplePropertyHover() throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-with-multiple-properties.xml", languageService);
+		Position pos = new Position(16, 12);
+		Hover hover = languageService.doHover(document,pos, new SharedSettings());
+		Range firstHoverRange = hover.getRange();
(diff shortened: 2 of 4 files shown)


lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -191,6 +191,29 @@ public void testPropertyDefinitionSameDocument() throws IOException, Interrupted
...
+ 	public void testMultiplePropertyHover() throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-with-multiple-properties.xml", languageService);
+		Position pos = new Position(16, 12);
+		Hover hover = languageService.doHover(document,pos, new SharedSettings());
+		System.out.println((((MarkupContent) hover.getContents().getRight()).getValue()));
+		System.out.println(hover.getRange());
+		Range firstHoverRange = hover.getRange();
+		
+
+		//DOMNode propertyNode = DOMUtils.findNodesByLocalName(document, "myProperty").stream().filter(node -> node.getParentElement().getLocalName().equals("properties")).collect(Collectors.toList()).get(0);;
```
AObuchow, line 204: Remove

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -335,8 +339,20 @@ private Hover collectProperty(IPositionRequest request, String property) {
 			for (Entry<String, String> prop : allProps.entrySet()) {
 				String mavenProperty = prop.getKey();
 				if (property.equals(mavenProperty)) {
-					return new Hover(new MarkupContent("markdown", "Property: " + mavenProperty
+					Hover hover = new Hover(new MarkupContent("markdown", "Property: " + mavenProperty
 							+ MavenPluginUtils.LINE_BREAK + "Value: " + prop.getValue() + MavenPluginUtils.LINE_BREAK));
+					try {
+						Position propertyStart = new Position(request.getPosition().getLine(),
```
AObuchow, line 345: This is code copied from MavenDefinitionParticipant. It should be factored into a function

AObuchow:
@mickaelistria this commit now adds a range but the same bug still occurs when testing as a user. I'm going to investigate further.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -146,16 +152,18 @@ private Hover collectArtifactDescription(IHoverRequest request, boolean isPlugin
...
-				.map(file -> builder.buildRawModel(file, ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL, false).get())
-				.map(model -> model.getName() + "\n\n" + model.getDescription())
-				.map(message -> (message.length() > 2 ? message : null));
+					.filter(gav -> (artifactToSearch.getGroupId() == null
+							|| artifactToSearch.getGroupId().equals(gav.getGroupId()))
+							&& (artifactToSearch.getArtifactId() == null
+									|| artifactToSearch.getArtifactId().equals(gav.getArtifactId()))
+							&& (artifactToSearch.getVersion() == null
+									|| artifactToSearch.getVersion().equals(gav.getVersion())))
+					.sorted(Comparator.comparing((Gav gav) -> new DefaultArtifactVersion(gav.getVersion())).reversed())
```
AObuchow, line 161: Didn't intend to format all this..

AObuchow:
My current assumption is that an upstream fix is required for onText hover, but I have to check first

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
```diff
@@ -61,6 +64,29 @@ public static DOMNode findClosestParentNode(final IPositionRequest request, fina
...
+	 * node
+	 * 
+	 * @param request
+	 * @param text
+	 * @return the range of the specificed text within the IPositionRequest's node,
+	 *         or null if the text could not be found in the node.
+	 * @throws DOMException
+	 * @throws BadLocationException
+	 */
+	public static Range getTextRange(IPositionRequest request, String text) throws DOMException, BadLocationException {
```
mickaelistria, line 79: I believe this would fail in some case, for example if I use `<blah>${property} ${property}</blah>`, then a request on the 2nd property would result on the range of the 1st one (if I' not mistaken).
Maybe `getMavenPropertyInRequest` should instead return a `Pair<Range, String>` where the range would be the actual range under the request that match a property.

AObuchow:
I think the issue is that LemMinX is using [`mergeHover`](https://github.com/eclipse/lemminx/blob/374ea32776f24a910f6a2541d0f66dd21cff1266/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLHover.java#L248).

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
AObuchow: @mickaelistria I see the issue your mentioning and agree it should be fixed for this case. However I'm not sure I understand your proposed solution unfortunately. Could you give a more detailed explanation or example?

mickaelistria:
If I'm not mistaken, mergeHover only affects ranges when they're not pre-set. If LemMinX-Maven sets (more correct) ranges explicitly first, then mergeHover doesn't change them.

AObuchow:
> If I'm not mistaken, mergeHover only affects ranges when they're not pre-set. If LemMinX-Maven sets (more correct) ranges explicitly first, then mergeHover doesn't change them.

```Java
		// Find the best hover range
		Range range = defaultRange;
		for (Hover hover : hovers) {
			if (hover.getRange() != null) {
				if (range == null) {
					range = hover.getRange();
				} else {
					// TODO : compute the best range
				}
			}
		}
		return MarkupContentFactory.createHover(contents, range);
```
If multiple hovers are present, it seems to use the range from the first hover in the list of hovers.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/DOMUtils.java
mickaelistria: The operation that finds a property name for a given position is actually iterating on the string and this same iteration can be used to find the range. As it appears the range is important, and that just looking for a substring is too weak, then we should also return the range when returning the property name.

AObuchow:
Upon debugging, mergeHovers only receives 1 hover so that's not causing the issue.

mickaelistria:
@AObuchow is there any progress here?

AObuchow:
> @AObuchow is there any progress here?

Yes sorry for the delay, I will be submitting a new PR today.
I've implemented the changes requested and while doing so I think I realized where the bug was originating.
`collectProperty` in MavenHoverParticipant was calling `DOMUtils.getTextRange()` which only finds the first occurrence range of a string in a node. It should instead be returning a list of occurrence ranges.

mickaelistria:
I think such `DOMUtils.getTextRange()` should simply doesn't exist at all, and a similar method should instead take as parameter the position, to prefer the instance of the string which is under given position.

AObuchow:
> I think such `DOMUtils.getTextRange()` should simply doesn't exist at all, and a similar method should instead take as parameter the position, to prefer the instance of the string which is under given position.

Agreed, I'll do this

AObuchow:
Confusingly, https://github.com/eclipse/lemminx-maven/issues/107 seems to be occurring again (I'm going to add a dedicated test which I should have already added), and the issue this PR addresses is not fixed.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/SimpleModelTest.java
```diff
@@ -191,6 +191,55 @@ public void testPropertyDefinitionSameDocument() throws IOException, Interrupted
...
+  		assertTrue((((MarkupContent) hover.getContents().getRight()).getValue().contains("test")));
+  		assertFalse(firstHoverRange.equals(secondHoverRange));
+	}
+	
+	@Test
+ 	public void testMultiplePropertyDefinitionRangeSameTag() throws IOException, InterruptedException, ExecutionException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-with-multiple-properties.xml", languageService);
+		Position pos = new Position(16, 12);
+		
+		List<? extends LocationLink> definitionLinks = languageService.findDefinition(document, pos, () -> {
```
AObuchow, line 214: Format this, remove println's

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -339,15 +344,26 @@ public static String getMavenPropertyInRequest(IPositionRequest request) {
...
 		if (indexOpen > indexCloseBefore) {
-			return tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+			Position propertyStart;
+			try {
+				String propertyText = tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+				propertyStart = new Position(request.getPosition().getLine(),
+						request.getXMLDocument().positionAt(request.getNode().getStart()).getCharacter() + indexOpen);
+				
+				// 3 is added to the end position's character to compensate for the '${}' 
+				Range propertyRange = new Range(propertyStart, new Position(request.getPosition().getLine(),
```
AObuchow, line 354: This works although it could probably be a bit more elegant, but shouldn't be a blocker

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -339,15 +344,26 @@ public static String getMavenPropertyInRequest(IPositionRequest request) {
 		int indexCloseBefore = beforeHoverText.lastIndexOf('}');
 		int indexCloseAfter = afterHoverText.indexOf('}');
 		if (indexOpen > indexCloseBefore) {
-			return tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+			Position propertyStart;
+			try {
+				String propertyText = tagText.substring(indexOpen + 2, indexCloseAfter + beforeHover);
+				propertyStart = new Position(request.getPosition().getLine(),
```
mickaelistria, line 350: What about just `positionStart = request.getXMLDocument().positionAt(request.getNode().getStart() + indexOpen)` ?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
mickaelistria: similarly
```java
int textStart = request.getNode().getStart();
Range propertyRange = XMLPositionUtility.createRange(textStart + indexOpen + 2, textStart + indexCloseAfter - 1, request.getXMLDocument());
```

AObuchow:
@mickaelistria this should be good to merge now

PR #122, opened 2020-07-29
Only use Markdown formatting when client supports it

pauldpong:
- Added a convertToBold function in MavenPluginUtils that allows for formatting based on if the hover request supports markdown. 
- Returned MarkupContent now has the correct MarkupKind 
- Added null checks/blank string checks to make hover clearer and no more 'null' strings

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lemminx.maven;
+
+public class MarkdownUtils {
+	public static final String LINE_BREAK = "\n\n";
+
+	private MarkdownUtils() {}
+	
+	public static String getLineBreak(boolean supportsMarkdown) {
+		return supportsMarkdown ? MarkdownUtils.LINE_BREAK : "\n";
+	}
+	
+	public static String toBold(String message) {
+		StringBuilder bold = new StringBuilder();
+
+		bold.append("**");
+		bold.append(message.trim());
+		bold.append("**");
+		
+		if (message.endsWith(" ")) {
+			bold.append(" ");
+		}
+		
+		return bold.toString();
+	}
+
+	public static String htmlXMLToMarkdown(String description) {
+		if (description.contains("<pre>") && description.contains("&lt;")) {
+			//Add markdown formatting to XML
+			String xmlContent = description.substring(description.indexOf("<pre>") + 6, description.indexOf("</pre>") - 1);
+			description = description.substring(0, description.indexOf("<pre>"));
+			xmlContent = xmlContent.replaceAll("&lt;", "<");
+			xmlContent = xmlContent.replaceAll("&gt;", ">");
+			xmlContent = "```XML" + "\n" + xmlContent + "\n" + "```";
+			description = description + LINE_BREAK + xmlContent;
+		}
+		return description;
+	}
+}
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
@@ -25,6 +25,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.repository.ArtifactRepository;
@@ -55,6 +56,7 @@ import org.eclipse.lemminx.services.extensions.IHoverRequest;
 import org.eclipse.lemminx.services.extensions.IPositionRequest;
 import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.MarkupContent;
+import org.eclipse.lsp4j.MarkupKind;
 
 public class MavenHoverParticipant implements IHoverParticipant {
 	private final MavenProjectCache cache;
@@ -130,6 +132,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 	}
 
 	private Hover collectArtifactDescription(IHoverRequest request, boolean isPlugin) {
+		boolean supportsMarkdown = request.canSupportMarkupKind(MarkupKind.MARKDOWN);
 		Collection<String> possibleHovers = Collections.synchronizedSet(new LinkedHashSet<>());
 		DOMNode node = request.getNode();
 		DOMDocument doc = request.getXMLDocument();
@@ -154,10 +157,24 @@ public class MavenHoverParticipant implements IHoverParticipant {
 				.findFirst()
 				.map(localRepoSearcher::findLocalFile)
 				.map(file -> builder.buildRawModel(file, ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL, false).get())
-				.map(model -> model.getName() + "\n\n" + model.getDescription())
+				.map(model -> {
+					UnaryOperator<String> toBold = supportsMarkdown ? MarkdownUtils::toBold : UnaryOperator.identity();
+					String lineBreak = MarkdownUtils.getLineBreak(supportsMarkdown);
+					String message = "";
+					
+					if (model.getName() != null) {
+						message += toBold.apply(model.getName());
+					}
+					
+					if (model.getDescription() != null) {
+						message += lineBreak + model.getDescription();
+					}
+					
+					return message;
+				})
 				.map(message -> (message.length() > 2 ? message : null));
 			if (localDescription.isPresent()) {
-				return new Hover(new MarkupContent("plaintext", localDescription.get()));
+				return new Hover(new MarkupContent(supportsMarkdown ? MarkupKind.MARKDOWN : MarkupKind.PLAINTEXT, localDescription.get()));
 			}
 		} catch (Exception e1) {
 			e1.printStackTrace();
@@ -194,7 +211,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 			return null;
 		}
 		
-		return new Hover(new MarkupContent("plaintext", possibleHovers.iterator().next()));
+		return new Hover(new MarkupContent(MarkupKind.PLAINTEXT, possibleHovers.iterator().next()));
 	}
 
 	private Hover collectGoal(IPositionRequest request) {
@@ -204,7 +221,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 			pluginDescriptor = MavenPluginUtils.getContainingPluginDescriptor(request, cache, repoSession, pluginManager);
 			for (MojoDescriptor mojo : pluginDescriptor.getMojos()) {
 				if (!node.getNodeValue().trim().isEmpty() && node.getNodeValue().equals(mojo.getGoal())) {
-					return new Hover(new MarkupContent("plaintext", mojo.getDescription()));
+					return new Hover(new MarkupContent(MarkupKind.PLAINTEXT, mojo.getDescription()));
 				}
 			}
 		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
@@ -213,7 +230,8 @@ public class MavenHoverParticipant implements IHoverParticipant {
 		return null;
 	}
 
-	private Hover collectPluginConfiguration(IPositionRequest request) {
+	private Hover collectPluginConfiguration(IHoverRequest request) {
+		boolean supportsMarkdown = request.canSupportMarkupKind(MarkupKind.MARKDOWN);
 		Set<MojoParameter> parameters;
 		try {
 			parameters = MavenPluginUtils.collectPluginConfigurationMojoParameters(request, cache, repoSession, pluginManager, buildPluginManager, mavenSession);
@@ -227,7 +245,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 			// The configuration element being hovered is at the top level
 			for (MojoParameter parameter : parameters) {
 				if (node.getLocalName().equals(parameter.getName())) {
-					return new Hover(MavenPluginUtils.getMarkupDescription(parameter, null));
+					return new Hover(MavenPluginUtils.getMarkupDescription(parameter, null, supportsMarkdown));
 				}
 			}
 		}
@@ -248,7 +266,7 @@ public class MavenHoverParticipant implements IHoverParticipant {
 					MojoParameter nestedParameter = parentParameter.getNestedParameters().get(0);
 					Class<?> potentialInlineType = PlexusConfigHelper.getRawType(nestedParameter.getParamType());
 					if (potentialInlineType != null && PlexusConfigHelper.isInline(potentialInlineType)) {
-						return new Hover(MavenPluginUtils.getMarkupDescription(nestedParameter, parentParameter));
+						return new Hover(MavenPluginUtils.getMarkupDescription(nestedParameter, parentParameter, supportsMarkdown));
 					}
 				}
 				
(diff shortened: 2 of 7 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -42,18 +43,46 @@
...
+	public static MarkupContent getMarkupDescription(Parameter parameter, boolean supportsMarkdown) {
 		String description = parameter.getDescription();
 		description = htmlXMLToMarkdown(description);
-		return new MarkupContent("markdown",
-				"**required:** " + parameter.getRequirement() + LINE_BREAK + "**Type:** " + parameter.getType() + LINE_BREAK
-						+ "Expression: " + parameter.getExpression() + LINE_BREAK + "Default Value: " + parameter.getDefaultValue()
-						+ LINE_BREAK + description);
+		
+		String markdownDescription = 
+				convertToBold("Required: ", supportsMarkdown) + parameter.isRequired() + LINE_BREAK +
```
pauldpong, line 69: If someone could verify that I am calling the correct function here. Before it for the "required" tag we would call parameter.getRequirement but comparing this to the name of the tag and similar code in the function below, I guess that it is supposed to be parameter.getRequired.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -42,18 +43,46 @@
 	private MavenPluginUtils() {
 		// Utility class, not meant to be instantiated
 	}
+	
+	public static String convertToBold(String message, boolean supportsMarkdown) {
```
mickaelistria, line 47: It looks like this method does more or less nothing when supportsMarkdown=false.
I suggest we just don't call it in such case.
Also, such methods would better fit in a new MarkdownUtils class or something like that.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
```diff
@@ -0,0 +1,19 @@
+package org.eclipse.lemminx.maven;
```
mickaelistria, line 1: Missing copyright header.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -154,10 +157,23 @@ private Hover collectArtifactDescription(IHoverRequest request, boolean isPlugin
 				.findFirst()
 				.map(localRepoSearcher::findLocalFile)
 				.map(file -> builder.buildRawModel(file, ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL, false).get())
-				.map(model -> model.getName() + "\n\n" + model.getDescription())
+				.map(model -> {
+					UnaryOperator<String> toBold = message -> supportsMarkdown ? MarkdownUtils.toBold(message) : message;
```
mickaelistria, line 161: This can probably be replaced by `UnaryOperator<String> toBold = supportsMarkdown ? MarkdownUtils::toBold : UnaryOperator.identity()`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -326,17 +343,23 @@ public static String getMavenPropertyInRequest(IPositionRequest request) {
...
 	}
 
-	private Hover collectProperty(IPositionRequest request, String property) {
+	private Hover collectProperty(IHoverRequest request, String property) {
+		boolean supportsMarkdown = request.canSupportMarkupKind(MarkupKind.MARKDOWN);
 		DOMDocument doc = request.getXMLDocument();
 		MavenProject project = cache.getLastSuccessfulMavenProject(doc);
 		if (project != null) {
 			Map<String, String> allProps = getMavenProjectProperties(project);
+			UnaryOperator<String> toBold = message -> supportsMarkdown ? MarkdownUtils.toBold(message) : message;
```
mickaelistria, line 352: This can probably be replaced by `UnaryOperator<String> toBold = supportsMarkdown ? MarkdownUtils::toBold : UnaryOperator.identity()`

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
```diff
@@ -326,17 +343,23 @@ public static String getMavenPropertyInRequest(IPositionRequest request) {
...
 
 			for (Entry<String, String> prop : allProps.entrySet()) {
 				String mavenProperty = prop.getKey();
 				if (property.equals(mavenProperty)) {
-					return new Hover(new MarkupContent("markdown", "Property: " + mavenProperty
-							+ MavenPluginUtils.LINE_BREAK + "Value: " + prop.getValue() + MavenPluginUtils.LINE_BREAK));
+					String message = toBold.apply("Property: ") + mavenProperty 
+							+ MavenPluginUtils.LINE_BREAK 
+							+ toBold.apply("Property: ") + prop.getValue() 
+							+ MavenPluginUtils.LINE_BREAK;
```
AObuchow, line 360: `MavenPluginUtils.LINE_BREAK` could probably be moved to `MarkdownUtils`

AObuchow: Also, the message remains the same regardless of whether markdown is used. I'm not sure how much of an issue this is, but since MavenPluginUtils.LINE_BREAK (which is "/n/n") is used, there will be double new lines in the plaintext message. 

Maybe something like:
```Java
String lineBreak = supportsMarkdown? MavenPluginUtils.LINE_BREAK : "/n"
```
would work?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
pauldpong: Since the identity function returns an Object, should I just cast it to a String?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -43,17 +45,30 @@ private MavenPluginUtils() {
...
+		UnaryOperator<String> toBold = message -> supportsMarkdown ? MarkdownUtils.toBold(message) : message;
 		String description = parameter.getDescription();
 		description = htmlXMLToMarkdown(description);
-		return new MarkupContent("markdown",
-				"**required:** " + parameter.getRequirement() + LINE_BREAK + "**Type:** " + parameter.getType() + LINE_BREAK
-						+ "Expression: " + parameter.getExpression() + LINE_BREAK + "Default Value: " + parameter.getDefaultValue()
-						+ LINE_BREAK + description);
+		
+		String markdownDescription = 
+				toBold.apply("Required: ") + parameter.isRequired() + LINE_BREAK +
```
AObuchow, line 54: My previous mention about LINE_BREAK not depending on whether plaintext or markdown is used also applies here. @mickaelistria WDYT? Should lineBreak be a variable dependant on supportsMarkdown?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
mickaelistria: identity() returns a `UnaryOperator<T>` so the type should be properly inferred by the compiler to `UnaryOperator<String>`.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
mickaelistria: yes.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
AObuchow: both methods (parameter.isRequired() and parameter.getRequirement()) sound like they return the same value, maybe inspect their implementations to see if there are any differences?

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenHoverParticipant.java
pauldpong: When I change it, I get a ```Type mismatch: cannot convert from UnaryOperator<Object> to String```. Should I just ignore it?

mickaelistria: This works for me locally. Are you using the exact same syntax I mentioned earlier?

pauldpong: So sorry, that's my bad, all working now.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
AObuchow: [Here's one](https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java#L1) you could copy and paste, for reference.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -34,26 +35,42 @@
...
-	static final String LINE_BREAK = "\n\n";
-
 	private MavenPluginUtils() {
 		// Utility class, not meant to be instantiated
 	}
 
-	public static MarkupContent getMarkupDescription(Parameter parameter) {
+	public static MarkupContent getMarkupDescription(Parameter parameter, boolean supportsMarkdown) {
+		UnaryOperator<String> toBold = supportsMarkdown ? MarkdownUtils::toBold : UnaryOperator.identity();
+		String lineBreak = supportsMarkdown ? MarkdownUtils.LINE_BREAK : "\n";
```
AObuchow, line 48: Might be a bit excessive, but this could be changed to something like `String lineBreak = MarkdownUtils.getLineBreak(supportsMarkdown)` to avoid duplication of `String lineBreak = supportsMarkdown ? MarkdownUtils.LINE_BREAK : "\n";`

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/PluginResolutionTest.java
```diff
@@ -67,9 +70,16 @@ public void testPluginConfigurationHoverMissingTransitiveDependency()
 			throws IOException, InterruptedException, ExecutionException, URISyntaxException {
 		assertFalse(initialMavenPluginApiDirectory.exists());
 		// <compilerArguments> hover
+		
+		SharedSettings settings = new SharedSettings();
```
AObuchow, line 74: @mickaelistria do you think a JUnit rule or something should be used here? To avoid duplicating the markdown hover settings

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
pauldpong: Yeah, don't think its too excessive, I'll change to that.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
pauldpong: They return different things, one is just a boolean if the tag is "required" and the other is the component's requirements. I'm guessing the correct one here is isRequired.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lemminx.maven;
+
+public class MarkdownUtils {
+	static final String LINE_BREAK = "\n\n";
```
mickaelistria, line 12: This one should probably better be public.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MarkdownUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+	private MarkdownUtils() {}
+	
+	public static String getLineBreak(boolean supportsMarkdown) {
```
mickaelistria, line 16: I don't think it fully belongs here.
Think for example of a case where we intentionally send HTML in case client is LSP4E because we know it works, then this methods becomes more a switch and wouldn't be only about markdown.
But it's not important and let's keep it at it for now. No need to over-engineer that now.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -34,26 +35,42 @@
...
+		
+		String markdownDescription = 
+				toBold.apply("Required: ") + parameter.isRequired() + lineBreak + 
+				toBold.apply("Type: ") + parameter.getType() + lineBreak;
+		
+		if (parameter.getExpression() != null && !parameter.getExpression().isEmpty()) {
+			markdownDescription += toBold.apply("Expression: ") + parameter.getExpression() + lineBreak; 
+		}
+		if (parameter.getDefaultValue() != null && !parameter.getDefaultValue().isEmpty()) {
+			markdownDescription += toBold.apply("Default Value: ") + parameter.getDefaultValue() + lineBreak;
```
mickaelistria, line 61: I think we should instead always show "Default Value: (unset)" in that case.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -62,22 +79,27 @@ public static MarkupContent getMarkupDescription(MojoParameter parameter, MojoPa
...
-				+ "Expression: " + expression + LINE_BREAK 
-				+ "Default Value: " + defaultValue + LINE_BREAK 
-				+ description;
-		// @formatter:on
-		return new MarkupContent("markdown", markdownDescription);
+				toBold.apply("Required: ") + parameter.isRequired() + lineBreak + 
+				toBold.apply("Type: ") + type + lineBreak;
+		
+		if (!expression.isEmpty()) {
+			markdownDescription += toBold.apply("Expression: ") + expression + lineBreak;
```
mickaelistria, line 93: Same as above, I believe "Expression: (none)" is useful to show.

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -62,22 +79,27 @@ public static MarkupContent getMarkupDescription(MojoParameter parameter, MojoPa
...
-		return new MarkupContent("markdown", markdownDescription);
+				toBold.apply("Required: ") + parameter.isRequired() + lineBreak + 
+				toBold.apply("Type: ") + type + lineBreak;
+		
+		if (!expression.isEmpty()) {
+			markdownDescription += toBold.apply("Expression: ") + expression + lineBreak; 
+		}
+		
+		if (!defaultValue.isEmpty()) {
+			markdownDescription += toBold.apply("Default Value: ") + defaultValue + lineBreak;
```
mickaelistria, line 97: "Default Value: (none)"

lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenPluginUtils.java
```diff
@@ -88,7 +110,7 @@ private static String htmlXMLToMarkdown(String description) {
 			xmlContent = xmlContent.replaceAll("&lt;", "<");
 			xmlContent = xmlContent.replaceAll("&gt;", ">");
 			xmlContent = "```XML" + "\n" + xmlContent + "\n" + "```";
-			description = description + LINE_BREAK + xmlContent;
+			description = description + MarkdownUtils.LINE_BREAK + xmlContent;
```
mickaelistria, line 113: This whole method should probably move to MarkdownUTils.

lemminx-maven/src/test/java/org/eclipse/lemminx/maven/test/PluginResolutionTest.java
mickaelistria: Maybe, but I don't think it's worth bothering too much about it.

pauldpong:
> See a few comments inline.
> @pauldpong If you prefer, we can merge this commit and iterate with more commits after, or we just keep improving the current commit, your call.

Let's keep it simple and keep improving the current commit.

AObuchow:
Awesome PR @pauldpong :) Merging!

Issue #134, opened 2020-09-02
Don't use System.out, System.err or e.printStackTrace


angelozerr:
Several code uses 

 * System.out (ex : https://github.com/eclipse/lemminx-maven/search?q=System.out&unscoped_q=System.out), 
 * System.err (ex: https://github.com/eclipse/lemminx-maven/search?q=System.err&unscoped_q=System.err) 
 * e.printStackTrace (ex: https://github.com/eclipse/lemminx-maven/search?q=printStackTrace&unscoped_q=printStackTrace)

If you wish to log something, I suggest that you use Java util logger API like we use for LemMinX (to be consistent).

ajo:
Hi, I've setup logging similar to what you have LemMinX, however test testDownloadArtifactOnHover fails. This test also appears to fail on a fresh pull. Is this related to issue #104, or an issue on my side?

AObuchow:
> Hi, I've setup logging similar to what you have LemMinX, however test testDownloadArtifactOnHover fails. This test also appears to fail on a fresh pull. Is this related to issue #104, or an issue on my side?

I believe it is related to #104 although I’m not certain yet. That test fails for everyone (including myself). When I wrote it, it was passing but perhaps it was only passing cause I had the artifact in my local repo already.

Overall, don’t worry if testDownloadArtifactOnHover is failing as it always seems to fail locally. This should be fixed at some point however, especially as it’s worrisome for newer contributors.

mickaelistria:
Fixed by @ajo with https://github.com/eclipse/lemminx-maven/pull/140 , thanks a lot!

PR #140, opened 2020-09-08
Issue #134 - Don't use System.out, System.err or e.printStackTrace

ajo:
Replace logging that used e.printStackTrace, System.err, and System.out with java.util.logging.Logger.


ajo:
Any idea what would be causing the tests to fail in Jenkins? I'm not able to reproduce locally. 

image

mickaelistria:
This looks all good, merging. Thanks a lot!
As for the failing test, it can be some network issue.

PR #156, opened 2020-10-04
Use a set to store tagOpen completions

### LINKED ISSUE #141: Duplicate dependency completion entries
Still trying to get lemminx-maven working in vscode.
Now I'm at a point where I get completion for dependencies, but I see duplicates for some of them:

Screen Shot 2020-09-08 at 7 22 21 PM

AObuchow:
Try to prevent duplicate completions
Fix #141

Signed-off-by: Andrew Obuchowicz <andrew@aobuchow.com>

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
@@ -179,6 +179,7 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 		if (request.getXMLDocument().getText().length() < 2) {
 			response.addCompletionItem(createMinimalPOMCompletionSnippet(request));
 		}
+		Set<CompletionItem> completionItems = new HashSet<CompletionItem>();
 		DOMElement parent = request.getParentElement();
 		if (parent == null || parent.getLocalName() == null) {
 			return;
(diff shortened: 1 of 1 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/maven/MavenCompletionParticipant.java
```diff
@@ -328,11 +329,12 @@ public void onXMLContent(ICompletionRequest request, ICompletionResponse respons
 					.stream()
 					.map(artifacts -> Collections.max(artifacts, highestVersionWithDescriptionComparator))
 					.map(artifactInfo -> toGAVCompletionItem(artifactInfo, request, gavInsertionStrategy))
-					.forEach(response::addCompletionItem);
+					.forEach(completionItems::add);
 		}
 		if (request.getNode().isText() && (allArtifactInfos.isEmpty() || request.getNode().getTextContent().contains("$"))) {
 			completeProperties(request).forEach(response::addCompletionAttribute);
```
AObuchow, line 335: I'm not sure what's the difference between a completion item and completion attribute? Maybe the properties should also be added to the set of completionItems.

mickaelistria:
Please add a test case beforehand for that issue, it will make things easier.

AObuchow:
> Please add a test case beforehand for that issue, it will make things easier.

I'm not entirely sure yet how to reproduce the bug as my suspicion is that the local repo & index were providing the same completion. Perhaps I can setup a test to download an artifact from the soffid repo (on hover) then trigger a completion to see if duplicate completions appear (one from local repo and one from index)

mickaelistria:
Yes, I also believe that local repo + index containing the artifact is the scenario. However, I don't think we should pick one or the other as in some cases, one or the other may have more data available. We should instead try to "reduce" multiple completion items for the same GAV to a single one, merging their input if its different.

AObuchow:
@mickaelistria ok good idea. So I'll try to come up with a function that merges GAV completion items to the one that has the most data available then.

mickaelistria:
This is too outdated, let's close.

PR #166, opened 2020-11-11
Migrate to Lemminx 0.14.1.

akurtakov:
Signed-off-by: Alexander Kurtakov <akurtako@redhat.com>


mickaelistria:
I think we can try to improve that deeper in tycho, by making the validate-version mojo m2e compatible so it would be run by default by m2e on build, and would report errors "naturally".
However, that wouldn't work with other IDEs, but I don't think we have to care, do we?

Issue #164, opened 2020-11-10
jacoco-maven-plugin cannot be resolved


rbioteau:
Hi !
When resolving the `jacoco-maven-plugin`, errors are thrown during the plugin resolution:

```
[ERROR] Failed to determine Java version for profile maven-jdk9 @ org.jacoco:org.jacoco.build:0.8.6
[ERROR] Failed to determine Java version for profile maven-jdk12 @ org.jacoco:org.jacoco.build:0.8.6
```
I guess it is because I do not have a jdk9 or jdk12 installed or configured in toolchains.
But the root cause is not displayed in the warning logged in the Problem view. I only have this misleading message:

```
Plugin could not be resolved. Ensure the plugin's groupId, artifactId and version are present.

Additional information: Plugin org.jacoco:jacoco-maven-plugin:0.8.6 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.jacoco:jacoco-maven-plugin:jar:0.8.6
```
Should the `validatePluginResolution` ignore those kind of errors ? By the way, the language server completion does not work properly for this plugin.

Full stacktrace:
```
org.apache.maven.plugin.PluginResolutionException: Plugin org.jacoco:jacoco-maven-plugin:0.8.6 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.jacoco:jacoco-maven-plugin:jar:0.8.6
	at org.apache.maven.plugin.internal.DefaultPluginDependenciesResolver.resolve(DefaultPluginDependenciesResolver.java:120)
	at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getPluginDescriptor(DefaultMavenPluginManager.java:182)
	at org.eclipse.lemminx.maven.MavenPluginUtils.getContainingPluginDescriptor(MavenPluginUtils.java:192)
	at org.eclipse.lemminx.maven.PluginValidator.validatePluginResolution(PluginValidator.java:49)
	at org.eclipse.lemminx.maven.PluginValidator.validateGoal(PluginValidator.java:112)
	at org.eclipse.lemminx.maven.MavenDiagnosticParticipant.doDiagnostics(MavenDiagnosticParticipant.java:73)
	at org.eclipse.lemminx.services.XMLDiagnostics.doExtensionsDiagnostics(XMLDiagnostics.java:85)
	at org.eclipse.lemminx.services.XMLDiagnostics.doDiagnostics(XMLDiagnostics.java:53)
[...]

PR #169, opened 2020-11-23
fix(plugin): resolution fails when plugin declared jdk profiles are not found

rbioteau:
See https://github.com/eclipse/lemminx-maven/issues/164


rbioteau:
Fixed by https://github.com/eclipse/lemminx-maven/commit/d626869e57787166008d2a098524e408a0fc3dbd

PR #182, opened 2021-01-08
Fix artifact completion in <dependencies> in VSCode

### LINKED ISSUE #180: Completion for partially complete artifact names in dependencies
Currently, you can add a dependency by opening completion at the `|`, typing the artifact name, then selecting which one you want:
```xml
...
  <dependencies>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <version>2.11.3</version>
    </dependency>
    |
  </depdenencies>
...
```

It would be helpful if completion could also be provided if you type part of the artifact name, then open completion (at the `|`):
```xml
...
  <dependencies>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <version>2.11.3</version>
    </dependency>
    junit-|
  </depdenencies>
...
```

It would also be helpful if completion could be provided inside of `<dependencies>` before any other dependencies are declared:
```xml
...
  <dependencies>
    junit-|
  </depdenencies>
...
```

datho7561:
Adjust the replace range of the TextEdit of the completion item so that it covers just the text node. This prevents VSCode from filtering the completion result.

Closes #180

Signed-off-by: David Thompson <davthomp@redhat.com>

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenCompletionParticipant.java
@@ -33,6 +33,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -75,6 +76,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either;
 public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 
 	private static final Logger LOGGER = Logger.getLogger(MavenCompletionParticipant.class.getName());
+	private static final Pattern ARTIFACT_ID_PATTERN = Pattern.compile("[-.a-zA-Z0-9]+");
 
 	static interface GAVInsertionStrategy {
 		/**
@@ -320,8 +322,10 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 						}).forEach(response::addCompletionItem);
 		}
 		if (!allArtifactInfos.isEmpty()) {
-			Comparator<ArtifactInfo> artifactInfoComparator = Comparator.comparing(artifact -> new DefaultArtifactVersion(artifact.getVersion()))/*.thenComparing(ArtifactInfo::getDescription)*/;
-			final Comparator<ArtifactInfo> highestVersionWithDescriptionComparator = artifactInfoComparator.thenComparing(artifactInfo -> artifactInfo.getDescription() != null ? artifactInfo.getDescription() : "");
+			Comparator<ArtifactInfo> artifactInfoComparator = Comparator.comparing(artifact -> new DefaultArtifactVersion(artifact.getVersion()));
+			final Comparator<ArtifactInfo> highestVersionWithDescriptionComparator = artifactInfoComparator
+					.thenComparing(
+							artifactInfo -> artifactInfo.getDescription() != null ? artifactInfo.getDescription() : "");
 			allArtifactInfos.stream()
 					.collect(Collectors.groupingBy(artifact -> artifact.getGroupId() + ":" + artifact.getArtifactId()))
 					.values()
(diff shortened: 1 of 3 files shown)


datho7561:
- [ ] Tests
- [ ] Take a look for regressions, for instance in Eclipse

angelozerr:
Thanks @datho7561 for your PR. For the test, it should be nice to consume LemMinx XMLAssert.

mickaelistria:
@datho7561 Thanks for the move to JUnit 5! Do you think you could make it a separate dedicated PR; we could merge it almost immediately.

datho7561:
@mickaelistria #187

mickaelistria:
@datho7561 Would it be possible to stick to LemMinX 0.14.1 for this change or are there API changes requiring the move to 0.14.2-SNAPSHOT? We may want to release LemMinX-Maven after a 1st batch of issue to have it working in VSCode is resolved; and we need to keep things compatible with Eclipse IDE, which at the moment is using 0.14.1 and may not be able to update to 0.14.2 as soon as it's released.

datho7561:
> ...are there API changes requiring the move to 0.14.2-SNAPSHOT? ...

There aren't API changes that are required, sorry I copy/pasted the dependency info from somewhere else and haven't fixed it in this branch yet.

> We may want to release LemMinX-Maven after a 1st batch of issue to have it working in VSCode is resolved; and we need to keep things compatible with Eclipse IDE, which at the moment is using 0.14.1 and may not be able to update to 0.14.2 as soon as it's released.

Okay sounds good.

mickaelistria:
I'll try it tomorrow morning in Eclipse IDE and will merge if I don't see regressions.

datho7561:
Okay. The test that I added fails as of right now.

mickaelistria:
So I'll wait for the test to pass before trying it in Eclipse IDE.
Do you have any clue why the test seems to not be run with https://github.com/eclipse/lemminx-maven/pull/182/checks?check_run_id=1683898573 while it's running with Jenkins?

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenCompletionParticipant.java
```diff
@@ -432,16 +434,34 @@ private CompletionItem toGAVCompletionItem(ArtifactInfo artifactInfo, ICompletio
 		if (artifactInfo.getDescription() != null) {
 			item.setDocumentation(artifactInfo.getDescription());
 		}
+		Range newReplaceRange;
+		try {
+			DOMDocument doc = request.getXMLDocument();
```
angelozerr, line 439: In LemMinx we had a similar usecase when you try to open completion for element tag name inside a text node which doesn't start with `<`. See https://github.com/eclipse/lemminx/blob/0d812fb4acca54b9df17371e0bb8e8b99bcf5b73/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java#L755

If you see `getElementNameRangeAt `https://github.com/eclipse/lemminx/blob/0d812fb4acca54b9df17371e0bb8e8b99bcf5b73/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java#L125 it consumes getWordRangeAt which use a regexp https://github.com/eclipse/lemminx/blob/0d812fb4acca54b9df17371e0bb8e8b99bcf5b73/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocument.java#L95

I suggest that you write a regexp pattern for artifact id and you do 

`Range artifactIdRange = doc.getTextDocument(request.getOffset(), ARTIFACTID_PATTERN);`

(artifactdIdRange is your newReplaceRange )

datho7561: Thanks for pointing this out! I have switched to this method for the PR

datho7561:
> So I'll wait for the test to pass before trying it in Eclipse IDE.

They should be good now.

> Do you have any clue why the test seems to not be run with https://github.com/eclipse/lemminx-maven/pull/182/checks?check_run_id=1683898573 while it's running with Jenkins?

I'll take a look into it.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/test/LemminxMavenCompletionParticipantTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static org.eclipse.lemminx.XMLAssert.c;
+import static org.eclipse.lemminx.XMLAssert.te;
+
+import org.eclipse.lemminx.XMLAssert;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(NoMavenCentralIndexExtension.class)
+public class LemminxMavenCompletionParticipantTest {
```
angelozerr, line 19: I would rename the class with just `MavenCompletionTest`

angelozerr:
Test on Windows OS with vscode and it works like a charm, great job @datho7561 !

mickaelistria:
Tested on m2e and seems to work as expected, no regression found.
Thanks @datho7561 !

Issue #181, opened 2021-01-08
java.nio.channels.OverlappingFileLockException with multiple instances trying to access same index


angelozerr:
Sometimes I have a NPE which comes from that `e.getCause()` at https://github.com/eclipse/lemminx-maven/blob/07361e1f0df7fd304c2a717bf14887331120e576/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/RemoteRepositoryIndexSearcher.java#L100 is null.

The real exception is `java.nio.channels.OverlappingFileLockException`:

![image](https://user-images.githubusercontent.com/1932211/104038893-f4b02300-51d5-11eb-98cc-e110ede49140.png)

mickaelistria:
This does usually happen when multiple instance of the language server are started. The issue is that the index files are protected by some lock; which is not a bad thing; but unfortunately it makes even just reading the index from different process non functioning. I tried to look at solutions some time ago, but didn't find anything useful. Any hint would be welcome!

mickaelistria:
Index stuff was removed with #250

PR #183, opened 2021-01-09
Fix NE when 'java.nio.channels.OverlappingFileLockException' occurs

angelozerr:
Fix NE when 'java.nio.channels.OverlappingFileLockException' occurs

See https://github.com/eclipse/lemminx-maven/issues/181


mickaelistria:
Thanks!

PR #190, opened 2021-01-15
Clean package names.

angelozerr:
Clean package names.


angelozerr:
This PR group features in a proper package name. It follows the same package names than other internal lemminx extension (with participants package name).

mickaelistria:
Seems like Snippets are broken by that change. Feel free to just remove anything snippet related by the way, we don't really need them IMO.

angelozerr:
> Seems like Snippets are broken by that change. Feel free to just remove anything snippet related by the way, we don't really need them IMO.

I removed it, but I would like to keep it since vscode-maven provides this kind of snippet. I will create an another PR for that to support this snippet but with the LemMinx snippet support which is more robust.

mickaelistria:
Thanks!

PR #191, opened 2021-01-16
Declare tag element in a DOM constants.

angelozerr:
Declare tag element in a DOM constants.


mickaelistria:
Thanks!

PR #192, opened 2021-01-18
launch for vscode debugging

angelozerr:
launch for vscode debugging


angelozerr:
This PR provides a vscode launch for debugging lemminx-maven inside vscode.

See https://github.com/redhat-developer/vscode-xml/blob/master/docs/Extensions.md#debugging-your-java-lemminx-extension for more explanation

mickaelistria:
Shouldn't that be part of vscode-xml instead?

angelozerr:
> Shouldn't that be part of vscode-xml instead?

The idea is that you import lemminx-maven in vscode, and you can debug it. We have a similar file in lemminx https://github.com/eclipse/lemminx/tree/master/.vscode

mickaelistria:
Why is that necessary for VSCode while this kind of thing is not necessary for other IDEs?

angelozerr:
It's just a launch (like an Eclipse IDE launch) which is proper configured.

mickaelistria:
Yes, but debugging Maven projects in Eclipse IDE usually work well without adding an Eclipse .launch file, and same for NetBeans and IntelliJ, I don't get what's so different with VSCode that makes it useful.

angelozerr:
> Yes, but debugging Maven projects in Eclipse IDE usually work well without adding an Eclipse .launch file, and same for NetBeans and IntelliJ, I don't get what's so different with VSCode that makes it useful.

In Eclipse IDE, you have an UI which creates the launch, in vscode you have not that and you must write at hand the launch. This PR avoid writting at hand this file. You import the lemminx-maven in vscode and you benefit from this launch without writing at hand.

mickaelistria:
> in vscode you have not that and you must write at hand the launch. 

OK, I thought there were some command in VSCode-Java to generate that kind of things.
I'll merge it because it doesn't harm and it makes you and others more productive; but I'm surprised that people have asked other IDEs a lot of effort to not require to check-in IDE-specific files in the project yet remain productive, and all of a sudden, committing another-IDE-specific file become almost necessary... Is this innovation or regression?

Issue #146, opened 2020-09-10
completion of properties without ${ prefix


a1dutch:
it would be good properties could be completed without having to add the `${` prefix

given the following property

```xml
<properties>
  <spring.version>5.0.2</spring.version>
<properties>
```

it would be good offer `${spring.version}` as a completion for below

```xml
<version>spring.ver</version>
```

mickaelistria:
That's a good idea, and it'd be relatively easyto write a test case for it and then to add the feature. Would you be interested in trying to write a patch?

a1dutch:
i can give it a go

mickaelistria:
Cool! Feel free to ask any question here if you're stuck.

mickaelistria:
@a1dutch Did you have the opportunity to look at it? If not, do you have plans for it?

a1dutch:
@mickaelistria i did start to have to have a look but the whole setup process to get it working in eclipse was far from easy

mickaelistria:
What specifically do you find complex importing it in Eclipse IDE? Just doing File > Open Projects from Filesystem... should work. Then it's just a typical Maven project: write a test case for the new feature, verifies it fails and do the necessary changes until it pases.

a1dutch:
yes i managed to get that part working, but trying to get it working with m2e & wild web developer to see it working using a locally built lemminx-maven

mickaelistria:
So if you already get it working in a Unit Test, please submit it as a PR without further consideration ;)

a1dutch:
let me check where i got too, i've been pretty busy since i started looking at it

mickaelistria:
@a1dutch I captured some answer to the "how to try in m2e" story in https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.editor.lemminx/HOWTO-DEV.md , but anyway, we first need the feature to work and be covered by unit test before it's worth trying to integrate it in client IDEs (eg Eclipse m2e).

PR #195, opened 2021-01-26
completion of properties without ${ prefix #146

vrubezhny:
Fixes: #146


angelozerr:
@vrubezhny that's great that you are using XMLAssert from LemMinx to write tests. Any feedback are welcome to improve it. Thanks!

PR #197, opened 2021-02-01
Add support for Maven XML settings in the JSON model.

### LINKED ISSUE #193: Language clients using different user settings format
On the vscode side, a user might set a property such as`"xml.maven.index.skip": true` in their `settings.json` file. This gets transformed to `{ "settings" { "xml" : { "maven" : { "index" : { "skip" : true } } } } }` when sent over in the `initialize` request. Ultimately, the attempt to set the property will fail because LemminxMavenExtension doesn't treat each "dot" of a property as a separate json object. See https://github.com/eclipse/lemminx-maven/blob/master/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java#L148 

I haven't checked, but I assume that on the Eclipse side, property keys with dots, are not transformed in the same way, so the settings likely work.

Before changing anything, I just want to get a consensus on what the right approach should be. I don't think we can stop the VSCode ecosystem from doing this transformation as it seems to happen at their platform level.

rgrunber:
- Resolves #193
- Add support for properties prefixed with "xml.maven.*"

Signed-off-by: Roland Grunberg <rgrunber@redhat.com>


rgrunber:
- I could move the newly created classes into a separate settings subpackage.
- I had `XMLMavenGeneralSettings` extend `XMLGeneralClientSettings` so that it could be parsed from the top InitializeParams but given that the initiliation options are set once again in MavenLemminxExtension to exclude the non xml ones, maybe it's possible to avoid this.

mickaelistria:
Thanks!

PR #199, opened 2021-02-05
Fix #196: lemminx-maven completely ignore local settings

### LINKED ISSUE #196: lemminx-maven completely ignore local settings
Hello,

I am using version 1.17.0.20201203-1258 (as told by the Window > Help for POM Editor using LemMinX language server / org.eclipse.m2e.lemminx.feature.feature.group).

Because I override the default .m2 local repository and put a file instead (eg: echo "FILE" >"${USERPROFILE}/.m2/repository"), Lemminx Maven is failing because it uses the default repository which I don't want:

```
Non-resolvable import POM: Could not transfer artifact org.junit:junit-bom:pom:5.7.0 from/to central-mirror (https://jcenter.bintray.com/): %USERPROFILE%\.m2\repository\org\junit\junit-bom\5.7.0\aether-738fca6a-b95a-43a5-90dc-8ccff46afeb0-junit-bom-5.7.0.pom-in-progress (Le chemin d’accès spécifié est introuvable)
```

To ensure that no one tries to write in `%USERPROFILE%\.m2\repository`, I created a file instead of a directory - hence the error.

The problem however is that Lemminx don't take into account the default settings.xml which starts with:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<settings xmlns="http://maven.apache.org/SETTINGS/1.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0
                      http://maven.apache.org/xsd/settings-1.1.0.xsd">
  <localRepository>E:\apps\data\maven-local</localRepository>
   ...
</settings>
```

The settings are configured to default in the Eclipse Maven preferences and I've got a lot of dubious warnings:

``` 
Plugin could not be resolved. Ensure the plugin's groupId, artifactId and version are present.

Additional information: Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0
```

I think those are related: the Language Server is explicitly **ignoring** the settings.xml for whatever reason.

The issue #42 was closed without much fixes.

A m2.conf as stated in the #42 could be fine, but I don't see why I would have to do that : my settings is in the default location, and it is not even being read otherwise it would have read the repo path...

Is there a way for me to force it to use the good settings/localRepository ?

How can I diagnose ? (the Language Server logs does not seems to provide hints)

glhez:
Do not change the LocalRepositoryPath to the default RepositorySystem.defaultUserLocalRepository if there some existing
  path.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
@@ -181,7 +181,7 @@ public class MavenLemminxExtension implements IXMLExtension {
 				requestPopulator.populateFromSettings(mavenRequest, userSettings);
 			}
 		}
-		mavenRequest.setLocalRepositoryPath(RepositorySystem.defaultUserLocalRepository);
+
 		String localRepoProperty = System.getProperty("maven.repo.local");
 		if (localRepoProperty != null) {
 			mavenRequest.setLocalRepositoryPath(new File(localRepoProperty));
@@ -192,6 +192,10 @@ public class MavenLemminxExtension implements IXMLExtension {
 			if (candidate.isFile() && candidate.canRead()) {
 				mavenRequest.setLocalRepositoryPath(candidate);
 			}
+		}
+
+		if (mavenRequest.getLocalRepositoryPath() == null) {
+			mavenRequest.setLocalRepositoryPath(RepositorySystem.defaultUserLocalRepository);
 		}
 
 		RepositorySystem repositorySystem = container.lookup(RepositorySystem.class);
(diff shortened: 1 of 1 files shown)


mickaelistria:
You need to sign the Eclipse Contributor Agreement ( "Sign" link on the right column of https://accounts.eclipse.org/legal/eca/validation/59174 ) and then amend your patch with `git commit --amend --signoff` to add the Signed-Off-By footer and then `push --force` to your branch. Then automated ECA check should then be happy and I'll be allowed to merge this good patch.

glhez:
Done. I also had to reset the author since I use github email shadowing...

glhez:
@mickaelistria I fail to understand the test that fail. On Jenkins, it fails. On Eclipse or maven (all on Windows), it works. 
The trace does not help: there is no CacheResourceDownloadingException thrown there.

mickaelistria:
The failure is not related to this patch. It's a failure that can happen when because of network glitches.
The patch is good as it, let's merge.

mickaelistria:
Thank you!

PR #203, opened 2021-02-16
Fix the indexer exception issues

vrubezhny:
The change fixes a number of NPPE and other kind of excceptions that appear in build log output
when the Cantral Maven Repository is disabled, like:

```
Caused by: java.lang.IllegalArgumentException: context mustn't be null
	at org.eclipse.lemminx.extensions.maven.searcher.RemoteRepositoryIndexSearcher.lambda$updateIndex$3(RemoteRepositoryIndexSearcher.java:229)
```
and
```
Caused by: java.lang.IllegalArgumentException: context mustn't be null
	at org.eclipse.lemminx.extensions.maven.searcher.RemoteRepositoryIndexSearcher.lambda$updateIndex$3(RemoteRepositoryIndexSearcher.java:229)
```
as well as some other exceptions.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/RemoteRepositoryIndexSearcher.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/RemoteRepositoryIndexSearcher.java
@@ -113,14 +113,18 @@ public class RemoteRepositoryIndexSearcher {
 				return indexDownloadJobs.get(res).thenApply(theVoid -> context);
 			}
 			final IndexingContext context = initializeContext(repositoryUrl);
-			indexingContexts.put(repositoryUrl, context);
-			CompletableFuture<IndexingContext> future;
-			try {
-				future = updateIndex(context).thenApply(theVoid -> context);
-				indexDownloadJobs.put(context, future);
-				return future;
-			} catch (ComponentLookupException e) {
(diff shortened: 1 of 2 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/RemoteRepositoryIndexSearcher.java
```diff
@@ -113,14 +113,19 @@ public RemoteRepositoryIndexSearcher(MavenLemminxExtension lemminxMavenPlugin, P
...
-				return CompletableFuture.failedFuture(e);
+			if (context != null) {
+				indexingContexts.put(repositoryUrl, context);
+				CompletableFuture<IndexingContext> future;
+				try {
+					future = updateIndex(context).thenApply(theVoid -> context);
+					indexDownloadJobs.put(context, future);
+					return future;
+				} catch (ComponentLookupException e) {
+					System.out.println("getIndexingContext(): logging out exeption: " + e.getLocalizedMessage());
```
mickaelistria, line 124: Please use a LOGGER here.

vrubezhny: This trace was added for debug purpose so, I'd better remove this line.

mickaelistria:
Thanks for this good patch!

Issue #77, opened 2020-06-02
Add support for dependency (validation, completion) on workspace projects


gayanper:
- Create a new module ModuleA in workspace
- Create another module ModuleB in the same workspace
- Now try to add a dependency in the ModuleB pom file to ModuleA, The dependency is not part of the completion list.

AObuchow:
There might be some useful code in `org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.loadPom(RepositorySystemSession, ArtifactDescriptorRequest, ArtifactDescriptorResult)` for this.

AObuchow:
Might be relevant:
- https://maven.apache.org/components/resolver-archives/resolver-LATEST/apidocs/org/eclipse/aether/DefaultRepositorySystemSession.html#setWorkspaceReader-org.eclipse.aether.repository.WorkspaceReader-
- https://maven.apache.org/components/resolver-archives/resolver-LATEST/apidocs/org/eclipse/aether/repository/WorkspaceReader.html

mickaelistria:
Adding validation to the scope of this bug. Eg create a parent pom that's not in a remote repo; reference it in another pom (without project.relativePath), then LemMinX-Maven doesn't resolve the parent to the other known pom.

mickaelistria:
This should now be fixed with merged patches.

We'll most likely face same performance issues as m2e on https://bugs.eclipse.org/bugs/show_bug.cgi?id=515668 with LemMinX-Maven about multiple project hierarchies stroed in memory, and the same approach (grouping project builds) will probably need to be used.
It won't be an easy task, it may actually a bit simpler for LemMinX-Maven than for m2e because things are better isolated than in m2e. However, we may also need to implement the "Facade" approach of m2e to simply avoid loading project; that would be a bigger piece of work!
But this is not currently highest priority for LemMinX-Maven.

PR #204, opened 2021-02-19
Add support for dependency (validation, completion) on workspace proj…

vrubezhny:
…ects #77

Content Assistant on Workspace project arfifacts


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
```diff
@@ -860,4 +863,36 @@ private static CompletionItem toCompletionItem(String label, String description,
...
+	
+	private void internalCollectWorkspaceArtifacts(ICompletionRequest request, Collection<ArtifactInfo> artifactInfosCollector,
+			Optional<String> groupId, Optional<String> artifactId) {
+		DOMElement parent = request.getParentElement();
+		if (parent == null || parent.getLocalName() == null) {
+			return;
+		}
+		DOMElement grandParent = parent.getParentElement();
+		if (grandParent == null || 
+				(!!PARENT_ELT.equals(grandParent.getLocalName()) &&
```
mickaelistria, line 875: `!!` seems like a bug.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
```diff
@@ -860,4 +863,36 @@ private static CompletionItem toCompletionItem(String label, String description,
 		Map<Object, Boolean> map = new ConcurrentHashMap<>();
 		return t -> map.putIfAbsent(keyExtractor.apply(t), Boolean.TRUE) == null;
 	}
+	
+	private void internalCollectWorkspaceArtifacts(ICompletionRequest request, Collection<ArtifactInfo> artifactInfosCollector,
+			Optional<String> groupId, Optional<String> artifactId) {
+		DOMElement parent = request.getParentElement();
```
mickaelistria, line 869: I don't get why line 869..879 are necessary.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
vrubezhny: Yes, For sure it is

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
vrubezhny: We need to know what is the parent node for proper filtering
We need to know what is the grand parent in order to calculate the only when we're in `<parent/> `, `<dependency/>` and `<plugin/>` nodes, if we remove these lines the proposals will be calculated for any `<groupId/>`, `<artifactID/>` and `<version/>` node contents encountered anywhere iside the document, for example inside `<project/>` node.

mickaelistria: OK, but shouldn't this filter be happening in the caller method? This kind of filtering doesn't seem to be specific to workspace completion, they're also true for local repo and index completion, aren't they?

vrubezhny: There is actually only `<parent/>`/`<plugin/>` separation in the caller method to filter the proposals, not sure if we need to replace it with any kind of filtering that will include workspace proposals.
I believe workspace proposals are to be added into the list of proposals found in local/remote/etc. repository searchers, not replacing them.
So placing parent/dependency/plugin filter here (inside `internalCollectWorkspaceArtifacts` makes `onXMLContent` method to be a bit shorter by preventing the copying of this filter among the three different cases of the caller's switch.

mickaelistria:
OK, let's merge as it and refactor later if needed.

PR #205, opened 2021-02-24
Add support for dependency (validation, completion) on workspace proj…

vrubezhny:
…ects #77

didChangeWorkspaceFolders realization for Workspace Service

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
@@ -12,12 +12,22 @@
 package org.eclipse.lemminx.extensions.maven;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import org.apache.maven.Maven;
 import org.apache.maven.artifact.repository.ArtifactRepository;
@@ -31,6 +41,7 @@ import org.apache.maven.execution.MavenSession;
 import org.apache.maven.internal.aether.DefaultRepositorySystemSessionFactory;
 import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.MavenPluginManager;
+import org.apache.maven.project.MavenProject;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.settings.Settings;
 import org.apache.maven.settings.io.SettingsReader;
@@ -61,6 +72,7 @@ import org.eclipse.lemminx.services.extensions.save.ISaveContext;
 import org.eclipse.lemminx.services.extensions.save.ISaveContext.SaveContextType;
 import org.eclipse.lemminx.settings.AllXMLSettings;
 import org.eclipse.lemminx.settings.InitializationOptionsSettings;
+import org.eclipse.lemminx.uriresolver.URIResolverExtensionManager;
 import org.eclipse.lsp4j.InitializeParams;
 
 import com.google.common.base.Objects;
@@ -92,6 +104,7 @@ public class MavenLemminxExtension implements IXMLExtension {
 	private BuildPluginManager buildPluginManager;
 
 	XMLMavenSettings settings = new XMLMavenSettings();
+	private URIResolverExtensionManager resolverExtensionManager = new URIResolverExtensionManager();
 
 	@Override
 	public void doSave(ISaveContext context) {
@@ -215,8 +228,6 @@ public class MavenLemminxExtension implements IXMLExtension {
 		return mavenRequest;
 	}
 
-
-
 	private static File getFileFromOptions(String element, File defaults) {
 		if (element == null) {
 			return defaults;
@@ -291,16 +302,20 @@ public class MavenLemminxExtension implements IXMLExtension {
 
 	public static boolean match(DOMDocument document) {
 		try {
-			File file = new File(URI.create(document.getDocumentURI()));
-			return (file.getName().startsWith("pom") && file.getName().endsWith(".xml"))
-					|| file.getName().endsWith(Maven.POMv4);
+			return match(new File(URI.create(document.getDocumentURI())).toPath());
 		} catch (Exception ex) {
 			// usually because of not so tolerant Java URI API
 			LOGGER.log(Level.SEVERE, ex.getMessage(), ex);
 			return false;
 		}
 	}
-
+	
+	public static boolean match(Path file) {
+		String fileName = file != null ? file.getFileName().toString() : null;
+		return fileName != null && ((fileName.startsWith("pom") && fileName.endsWith(".xml"))
+				|| fileName.endsWith(Maven.POMv4));
+	}
+	
 	public MavenProjectCache getProjectCache() {
 		initialize();
 		return this.cache;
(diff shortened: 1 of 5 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -92,6 +110,7 @@
 	private BuildPluginManager buildPluginManager;
 
 	XMLMavenSettings settings = new XMLMavenSettings();
+	private XMLLanguageService languageService = new XMLLanguageService();
```
mickaelistria, line 113: XMLLanguageService shouldn't be instantiated in the extension.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -337,9 +354,79 @@ public MavenPluginManager getMavenPluginManager() {
...
+		
+		return projectsToAdd;
+	}
+	
+	private DOMDocument createDOMDocument(URI uri) throws IOException, URISyntaxException {
+		File file = new File(uri);
+		String contents = Files.readString(file.toPath());
+		return org.eclipse.lemminx.dom.DOMParser.getInstance()
+				.parse(new TextDocument(new TextDocumentItem(uri.toString(), "xml", 1, contents)), 
+						languageService.getResolverExtensionManager());
```
mickaelistria, line 429: As you don't need extensions at that stage, you can probably just use `new URIResolverExtensionManager()` so you don't need an XMLLanguageService.

mickaelistria (changes requested):

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: OK

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -337,9 +354,79 @@ public MavenPluginManager getMavenPluginManager() {
...
+					public FileVisitResult preVisitDirectory(Path file, BasicFileAttributes attrs) throws IOException {
+						if (file.getFileName().toString().charAt(0) == '.') {
+							return FileVisitResult.SKIP_SUBTREE;
+						}
+						return FileVisitResult.CONTINUE;
+					}
+
+					@Override
+					public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+						if (file.getFileName().toString().toLowerCase().endsWith(".xml")) {
```
mickaelistria, line 399: This should reuse the same logic as `match` method. Maybe having a `match(File)` method would be good here.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -337,9 +354,79 @@ public MavenPluginManager getMavenPluginManager() {
...
+				});
+			} catch (IOException e) {
+				LOGGER.log(Level.SEVERE, e.getMessage(), e);
+			}
+		});
+		
+		return projectsToAdd;
+	}
+	
+	private DOMDocument createDOMDocument(URI uri) throws IOException, URISyntaxException {
```
mickaelistria, line 424: Couldn't this be part of ProjectCache instead?

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
```diff
@@ -113,6 +113,25 @@ private void check(DOMDocument document) {
 		}
 	}
 
+	/**
+	 * Parses and adds a document for a given URI into the projects cache
+	 * 
+	 * @param uri An URI of a document to add
+	 * @param document A document to add
+	 */
+	public void addDocument(URI uri, DOMDocument document ) {
```
mickaelistria, line 122: As mentioned above, I don't think the document should be an argument of this method as we can fully deduce it from the uri.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: `visitFile()` uses this method for filtering purposes (in order to collect only xml files with `<project/>` root element. Also, once a document is already created here it  is passed to `MavenProjectCache.addDocument(URI, DOMDocument)` method.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
vrubezhny: As I wrote above - once the document is already created in `visitFile()` having this parameter here allows to reuse it instead of creating it once again by URI inside `addDocument()` method.

mickaelistria: What makes it necessary to create the document in visitFile() ?

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
mickaelistria: I think we should just rely on filename at this stage and not look at document content here.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: I would change it to `match (Path)` as visitFile() operates with Path

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: Yes, but I have moved that root element check to addDocument(), so if a document doesn't have `<project />` element then it's just not added to the cache. This way it works much faster.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
vrubezhny: The purpose of this was detection of pom xml documents. 
I have moved this detection to `MavenProjectCache.addDocument(URI)`

mickaelistria:
Thanks!

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -92,6 +104,7 @@
 	private BuildPluginManager buildPluginManager;
 
 	XMLMavenSettings settings = new XMLMavenSettings();
+	private URIResolverExtensionManager resolverExtensionManager = new URIResolverExtensionManager();
```
angelozerr, line 107: Don't instantiate URIResolverExtensionManger but get it in start method with 

```java
public void start(InitializeParams params, XMLExtensionsRegistry registry) {
	this.resolverExtensionManager  = registry.getResolverExtensionManager();
...
```

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
```diff
@@ -68,6 +76,7 @@
 
 	private final List<Consumer<MavenProject>> projectParsedListeners = new ArrayList<>();
 	private Properties systemProperties;
+	private URIResolverExtensionManager resolverExtensionManager = new URIResolverExtensionManager();
```
angelozerr, line 79: Don't instainciate resolverExtensionManager, but use the resolverExtensionManager  from MavenLemminxExtension getted in start method.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
```diff
@@ -235,4 +274,12 @@ public PlexusContainer getPlexusContainer() {
 	public Collection<MavenProject> getProjects() {
 		return projectCache.values();
 	}
+	
+	private DOMDocument createDOMDocument(URI uri) throws IOException, URISyntaxException {
```
angelozerr, line 278: Replace this method with https://github.com/eclipse/lemminx/blob/d27272d14db97cba7f22a22d7a7c7bc104d28d14/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/DOMUtils.java#L128

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/MavenProjectCacheTest.java
```diff
@@ -76,6 +88,63 @@ public void testParentChangeReflectedToChild()
 		}
 	}
 
+	@Test
+	public void testAddFolders_didChangeWorkspaceFolders() throws Exception {
+		XMLLanguageService languageService = new XMLLanguageService();
```
angelozerr, line 93: I suggest you that you:

 * test range of diagnostics
 * use XMLAssert#testDiagnosticsFor from LemMinx

See https://github.com/eclipse/lemminx/blob/d27272d14db97cba7f22a22d7a7c7bc104d28d14/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSyntaxDiagnosticsTest.java#L48 for instance.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/MavenProjectCacheTest.java
```diff
@@ -76,6 +88,63 @@ public void testParentChangeReflectedToChild()
...
+		IWorkspaceServiceParticipant workspaceService = languageService.getWorkspaceServiceParticipants().stream().filter(MavenWorkspaceService.class::isInstance).findAny().get();
+		assertNotNull(workspaceService);
+		
+		URI folderUri = getClass().getResource("/modules").toURI();
+		WorkspaceFolder wsFolder = new WorkspaceFolder(folderUri.toString());
+
+		// Test add folders
+		workspaceService.didChangeWorkspaceFolders(
```
angelozerr, line 101: I would write `languageService.getWorkspaceServiceParticipants().forech(p -> p.didChangeWorkspaceFolders`

vrubezhny: We're going to test one certain participant - `MavenWorkspaceService` here in order to invoke its `didChangeWorkspaceFolders()` API .method
IMHO, Including more participants into the processing potentially may create unexpected effects causing test execution problems, at the same time not adding any useful functionality and not being a subject to test.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: ok

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
vrubezhny: ok

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
vrubezhny: ok

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/MavenProjectCacheTest.java
vrubezhny: We're verifying that NO Error diagnostics appear on the document, while, as far as I understand, `testDiagnosticsFor` always verify presence of a certain Error diagnostics code - so it doesn't applicable for this case.

Or I just couldn't find the correct way to use it...
Possible error in our case could be:
```
Diagnostic [
  range = Range [
    start = Position [
      line = 8
      character = 10
    ]
    end = Position [
      line = 8
      character = 11
    ]
  ]
  severity = Error
  code = null
  source = null
  message = "Non-resolvable parent POM for org.test.modules:ModuleC:0.0.1-SNAPSHOT: Could not find artifact org.test.modules:ModuleA:pom:0.0.1-SNAPSHOT"
  tags = null
  relatedInformation = null
]
```
So, if I replace the existing `assertFalse` to `testDiagnosticsFor`, ie do the following change:
```
//		assertFalse(diagnostics.stream().anyMatch(diag -> (diag.getMessage().contains("ModuleA"))));
		
		URI uri = getClass().getResource("/modules/dependent/module-c-pom.xml").toURI();
		testDiagnosticsFor(doc.getTextDocument().getText(), null, null, uri.toString(),  d(8, 10, 8, 11, null));
``` 
then `testDiagnosticsFor` call doesn't detect the failure if it appears - it's just doesn't work for this case

vrubezhny:
I've created a PR according to @angelozerr 's comments: 
https://github.com/eclipse/lemminx-maven/pull/206

PR #209, opened 2021-03-09
Cancel

rgrunber:


mickaelistria:
Please also bump the minor version when moving to API incompatible LemMinX. Also, if LemMinX breaks API, it'd be better to bump the minor segment, not just the service/patch segment.

lemminx-maven/pom.xml
```diff
@@ -22,7 +22,7 @@
 		<lemminx.maven.indexDirectory>${project.build.directory}/test-index</lemminx.maven.indexDirectory>
 		<maven.repo.local>${settings.localRepository}</maven.repo.local>
 		<junit-jupiter.version>5.7.0</junit-jupiter.version>
-		<lemminx.version>0.15.0</lemminx.version>
+		<lemminx.version>0.15.1-SNAPSHOT</lemminx.version>
```
datho7561, line 25: Probably worthwhile to upgrade to LemMinX 0.16.2 now

datho7561 (changes requested):
I had to change the source and target level to 8 to get it to work with LemMinX. Some of the tests are failing.

lemminx-maven/pom.xml
rgrunber: Updated to 0.16.2

mickaelistria:
That looks all good to me, and to CI. Let's merge.
Thanks!

PR #222, opened 2021-08-29
Fix #221: Repositories from settings get replaced by the default repo…

### LINKED ISSUE #221: Repositories from settings get replaced by the default repository and m2e wrongly report errors
Hello,

I am using several remote repositories - four, and some being corporate - and m2e trigger warning about missing dependencies whenever I edit my POM file. 

The dependency is inside a corporate repository, whose content is of course not accessible in central. 
The dependency was already downloaded by mvn (using mvn clean install...) but somehow, m2e tries to download it again - perhaps to check if it is up to date.

1) In a previous PR/issue (#196), I had the same kind of problem with the local repo path, which was overridden by some default value. It failed because I created a file ~/.m2/repository
2) The repositories are overridden from settings by the "default": https://github.com/eclipse/lemminx-maven/blob/437daf64d6a50b7a7e1f6ebc670a276694522ed1/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java#L224 (this is on master branch)

The code should perhaps be something like:

```java
List<ArtifactRepository> repositories = mavenRequest.getRemoteRepositories()
if (repositories.isEmpty()) repositories.add(defaultRemoteRepositories);
List<ArtifactRepository> pluginRepositories = mavenRequest.getPluginArtifactRepositories()
if (pluginRepositories .isEmpty()) pluginRepositories .add(defaultRemoteRepositories);
```

However, I fail to understand why this code differ from whatever is done when importing the maven project ? (since I only have this error when editing the pom file).

(I can do a PR to apply the code above)

glhez:


mickaelistria:
The test failure are not related to this change, and the patch is good. Let's merge it.
Thanks a lot, I think it's actually going to fix a bunch of other bugs!

PR #229, opened 2021-09-07
As lemminx server prototype and example this plugin might better deploy a single JARs 

svanteschubert:
A single JAR representing a lemminx plugin (JAR extension) is easier to declare in the package.json than dozens of JAR.

1. Instead of using the ZIP assembly file here the default build-in file:
see https://maven.apache.org/plugins/maven-assembly-plugin/usage.html
2. Lemminx itself is taken as provided and is not being included in the deployment (as protogé of a Lemminx example this is likely the default that user - like me writing an own lemminx extension - would like to see.
3. Additional information as version and build-time etc. had been added to a new manifest section.

Hope you enjoy it!
Svante


angelozerr:
@zcjava I think this PR could interest you, could you try it an dgive us feedback please, thanks!

mickaelistria:
The jar-with-dependencies would need to be verified before we ship it. Many Maven related jars have some specific component instructions files (eg META-INF/services or plexus/component.xml...) that need to remain visible even after merging the jars, while merging the jars would make those files override each other and ultimately prevent from proper execution.
Also, with the current zip-with-dependencies and lemminx being *not* provided, the zip-with-dependencies contains everything to start the language server; even lemminx itself; so it's a good standalone delivery that some may want to use.
A dedicated jar meant to be a complete lemminx extension would better be a separate packaging.

zcjava:
the PR may be has problem. lemminx-maven-0.4.1-SNAPSHOT-jar-with-dependencies.jar no include class file of lemminx
not found main class 
image
image
i unzip dependencies.jar dotnot include lemminx jar

zcjava:
> The jar-with-dependencies would need to be verified before we ship it. Many Maven related jars have some specific component instructions files (eg META-INF/services or plexus/component.xml...) that need to remain visible even after merging the jars, while merging the jars would make those files override each other and ultimately prevent from proper execution.
> Also, with the current zip-with-dependencies and lemminx being _not_ provided, the zip-with-dependencies contains everything to start the language server; even lemminx itself; so it's a good standalone delivery that some may want to use.
> A dedicated jar meant to be a complete lemminx extension would better be a separate packaging.

yes , keep standalone for plug-in package . but not everyone use idea or eclipse suite. someone need hand-build it in other editor. maybe support choose be necessary. this is not bad idea.

svanteschubert:
Hello everyone,

I did some testing and it turns out that there is indeed is a problem, that I guessed was solved by the handlers implicitly
https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-container-descriptor-handlers.html

But when I tested today and took Lemminx back in by removing the <scope>provided</scope> in the dependencies and bundle it again, the org.eclipse.lemminx.services.extensions.IXMLExtension file was indeed overwritten as stated by you above.
The lemminx-maven extension was not listed.

As I found out that vscode-xml allows wildcards in their JAR references the bundling is no longer such a high requirement.
In addition, I am not sure how if this all JARs in one JAR bundling might cause any further problems with >=JDK 9 module concept.

I suggest closing this PR. Sorry for the noise, learned a lesson :-)

Have a great day!
Svante

PS: **Could you please point out how you build this Lemminx JAR extension as VScode extension for the marketplace? You add them to vscode-xml with the JARs referenced by wildcards in package.json?**

mickaelistria:
Closing the PR as suggested.

> Could you please point out how you build this Lemminx JAR extension as VScode extension for the marketplace?

As far as I know, no-one tried it so far; so there is no doc or example available. Contributions for that would be very welcome, either on LemMinX, LemMinX-Maven or vscode-xml.

angelozerr:
> PS: Could you please point out how you build this Lemminx JAR extension as VScode extension for the marketplace? You add them to vscode-xml with the JARs referenced by wildcards in package.json?

https://github.com/redhat-developer/vscode-xml/blob/master/CONTRIBUTING.md should help you I think.

PR #236, opened 2021-11-29
Maven dependencies are sometimes downloaded in home folder #383

vrubezhny:

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -44,9 +45,13 @@ import org.apache.maven.model.validation.ModelValidator;
 import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.MavenPluginManager;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.properties.internal.EnvironmentUtils;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.settings.Settings;
-import org.apache.maven.settings.io.SettingsReader;
+import org.apache.maven.settings.building.DefaultSettingsBuildingRequest;
+import org.apache.maven.settings.building.SettingsBuilder;
+import org.apache.maven.settings.building.SettingsBuildingException;
+import org.apache.maven.settings.building.SettingsBuildingRequest;
 import org.codehaus.plexus.ContainerConfiguration;
 import org.codehaus.plexus.DefaultContainerConfiguration;
 import org.codehaus.plexus.DefaultPlexusContainer;
@@ -56,6 +61,7 @@ import org.codehaus.plexus.PlexusContainerException;
 import org.codehaus.plexus.classworlds.ClassWorld;
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
 import org.codehaus.plexus.classworlds.realm.NoSuchRealmException;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.WorkspaceReader;
 import org.eclipse.lemminx.dom.DOMDocument;
@@ -107,7 +113,7 @@ public class MavenLemminxExtension implements IXMLExtension {
 	XMLMavenSettings settings = new XMLMavenSettings();
 	private URIResolverExtensionManager resolverExtensionManager;
 	private List<WorkspaceFolder> initialWorkspaceFolders = List.of();
-
+	
 	@Override
 	public void doSave(ISaveContext context) {
 		if (context.getType() == SaveContextType.SETTINGS) {
@@ -188,25 +194,31 @@ public class MavenLemminxExtension implements IXMLExtension {
 
 	private MavenExecutionRequest initMavenRequest(PlexusContainer container, XMLMavenSettings options) throws Exception {
 		MavenExecutionRequest mavenRequest = new DefaultMavenExecutionRequest();
-		SettingsReader reader = container.lookup(SettingsReader.class);
-		MavenExecutionRequestPopulator requestPopulator = container.lookup(MavenExecutionRequestPopulator.class);
+		Properties systemProperties = getSystemProperties();
+		mavenRequest.setSystemProperties(systemProperties);
 		{
-			final File globalSettingsFile = getFileFromOptions(options.getGlobalSettings(), SettingsXmlConfigurationProcessor.DEFAULT_GLOBAL_SETTINGS_FILE);
-			if (globalSettingsFile.canRead()) {
-				mavenRequest.setGlobalSettingsFile(globalSettingsFile);
-				Settings globalSettings = reader.read(globalSettingsFile, null);
-				requestPopulator.populateFromSettings(mavenRequest, globalSettings);
-			}
-		}
-		{
-			final File localSettingsFile = getFileFromOptions(options.getUserSettings(), SettingsXmlConfigurationProcessor.DEFAULT_USER_SETTINGS_FILE);
-			if (localSettingsFile.canRead()) {
-				mavenRequest.setUserSettingsFile(localSettingsFile);
-				Settings userSettings = reader.read(localSettingsFile, null);
-				requestPopulator.populateFromSettings(mavenRequest, userSettings);
-			}
-		}
+			final File globalSettingsFile = getFileFromOptions(options.getGlobalSettings(),
+					SettingsXmlConfigurationProcessor.DEFAULT_GLOBAL_SETTINGS_FILE);
+			final File localSettingsFile = getFileFromOptions(options.getUserSettings(),
+					SettingsXmlConfigurationProcessor.DEFAULT_USER_SETTINGS_FILE);
 
+			Settings settngs = buildSettings(container,
+		    		globalSettingsFile.canRead() ? globalSettingsFile : null,
+		    		localSettingsFile.canRead() ? localSettingsFile : null,
+		    		systemProperties);
+			
+		    if (settngs != null) {
+				if (globalSettingsFile.canRead()) {
+					mavenRequest.setGlobalSettingsFile(globalSettingsFile);
+				}
+				if (localSettingsFile.canRead()) {
+					mavenRequest.setUserSettingsFile(localSettingsFile);
+				}
+				MavenExecutionRequestPopulator requestPopulator = container.lookup(MavenExecutionRequestPopulator.class);
+				requestPopulator.populateFromSettings(mavenRequest, settngs);
+		    }
+		}
+		
 		String localRepoProperty = System.getProperty("maven.repo.local");
 		if (localRepoProperty != null) {
 			mavenRequest.setLocalRepositoryPath(new File(localRepoProperty));
@@ -229,20 +241,57 @@ public class MavenLemminxExtension implements IXMLExtension {
 		List<ArtifactRepository> defaultRemoteRepositories = Collections.singletonList(repositorySystem.createDefaultRemoteRepository());
 		mavenRequest.setRemoteRepositories(joinRemoteRepositories(mavenRequest.getRemoteRepositories(), defaultRemoteRepositories));
 		mavenRequest.setPluginArtifactRepositories(joinRemoteRepositories(mavenRequest.getPluginArtifactRepositories(), defaultRemoteRepositories));
-		mavenRequest.setSystemProperties(System.getProperties());
+		mavenRequest.setSystemProperties(systemProperties);
 		mavenRequest.setCacheNotFound(true);
 		mavenRequest.setCacheTransferError(true);
 		mavenRequest.setWorkspaceReader(new MavenLemminxWorkspaceReader(this));
 		return mavenRequest;
 	}
+
+	public Settings buildSettings(PlexusContainer container, File globalSettingsFile, File userSettingsFile, Properties systemProperties) {
+		SettingsBuildingRequest request = new DefaultSettingsBuildingRequest();
+		request.setGlobalSettingsFile(globalSettingsFile);
+		request.setUserSettingsFile(userSettingsFile != null ? userSettingsFile
+				: SettingsXmlConfigurationProcessor.DEFAULT_USER_SETTINGS_FILE);
+		request.setSystemProperties(systemProperties);
+		try {
+			return container.lookup(SettingsBuilder.class).build(request).getEffectiveSettings();
+		} catch (SettingsBuildingException | ComponentLookupException ex) {
+			return null;
+		}
+	}
 	
+	/*
+	 * Thread-safe properties copy implementation.
+	 * <p>
+	 * {@link Properties#entrySet()} iterator is not thread safe and fails with
+	 * {@link ConcurrentModificationException} if the source properties "is
+	 * structurally modified at any time after the iterator is created". The
+	 * solution is to use thread-safe {@link Properties#stringPropertyNames()}
+	 * enumerate and copy properties.
+	 *
+	 * @see https://bugs.eclipse.org/bugs/show_bug.cgi?id=440696
+	 */
+	private static Properties getSystemProperties() {
+		Properties systemProperties = new Properties();
+		EnvironmentUtils.addEnvVars(systemProperties);
+		Properties from = System.getProperties();
+		for (String key : from.stringPropertyNames()) {
+			String value = from.getProperty(key);
+			if (value != null) {
+				systemProperties.put(key, value);
+			}
+		}
+		return systemProperties;
+	}
+
 	private List<ArtifactRepository> joinRemoteRepositories(List<ArtifactRepository> a, List<ArtifactRepository> b) {
-	  if (a.isEmpty()) {
-	    return b;
-	  }
-	  List<ArtifactRepository> remotes = new ArrayList<>(a);
-	  remotes.addAll(b);
-	  return remotes;	  
+		if (a.isEmpty()) {
+			return b;
+		}
+		List<ArtifactRepository> remotes = new ArrayList<>(a);
+		remotes.addAll(b);
+		return remotes;
 	}
 
 	private static File getFileFromOptions(String element, File defaults) {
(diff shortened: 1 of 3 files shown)


mickaelistria:
That looks good, however some of this logic is also present in `MavenProjectCache#newProjectBuildingRequest()`. Can you please remove this logic from MavenProjectCache so it would use `mavenRequest.getSystemProperties()` in the building request directly instead?

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,69 @@
...
+			plugin.settings.setUserSettings(localSettingsContent);
+			
+			invoke(plugin, MavenLemminxExtension.class, "initialize", null, null);
+			
+			File localRepositoryPath = plugin.getMavenSession().getRequest().getLocalRepositoryPath();
+			assertNotNull(localRepositoryPath);
+			assertFalse(localRepositoryPath.getAbsolutePath().contains("${"));
+			assertTrue(localRepositoryPath.canRead());
+		} finally {
+			System.setProperty("maven.repo.local", localRepoProperty); // Restore the value
```
mickaelistria, line 48: this needs to verify localRepoProperty is not null, or it throws a NPE.

mickaelistria:
The test does pass when running locally and *without* the fix in MavenLemminxExtension from the IDE. Please make sure the test reveals the issue in such case.
Note that it would be interesting to use a "dummy" env var and a dummy path as here the given path matches the fallback strategies that are in place; so we cannot ensure that it's env-var expansion happening correctly, or whether it's the fallback strategies doing their job.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,69 @@
...
+		String localRepoProperty = System.getProperty("maven.repo.local");
+		try {
+			Properties properties = System.getProperties();
+			properties.remove("maven.repo.local");
+			System.setProperties(properties);	// Not to use local repository set from properties
+			MavenLemminxExtension plugin = new MavenLemminxExtension();
+			plugin.settings.setGlobalSettings(null);
+			plugin.settings.setUserSettings(localSettingsContent);
+			
+			invoke(plugin, MavenLemminxExtension.class, "initialize", null, null);
```
mickaelistria, line 41: this should be removed because `plugin.getMavenSession()` does invoke initialize.

mickaelistria (changes requested):

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,69 @@
...
+			
+			File localRepositoryPath = plugin.getMavenSession().getRequest().getLocalRepositoryPath();
+			assertNotNull(localRepositoryPath);
+			assertFalse(localRepositoryPath.getAbsolutePath().contains("${"));
```
mickaelistria, line 45: Test should verify the actual path is `System.getProperty("user.home") + ".m2/bug383repository"` instead of just checking for env-var expansion; so we know it's not a fallback strategy defalting to default repo (which would trigger a failure here).

mickaelistria (changes requested):

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: ACK

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: ACK

vrubezhny:
> The test does pass when running locally and _without_ the fix in MavenLemminxExtension from the IDE. Please make sure the test reveals the issue in such case. 

It does fail for me if I run the test using the old (_without_ the fix) MavenLemminxExtension and MavenProjectCache classes versions.
Does it really succeed for you?

> Note that it would be interesting to use a "dummy" env var and a dummy path as here the given path matches the fallback strategies that are in place; so we cannot ensure that it's env-var expansion happening correctly, or whether it's the fallback strategies doing their job.

I agree that using such a var would make the test results to be more "deterministic".
Going to realize it.

mickaelistria:
> Does it really succeed for you?

Yes, although when looking at the path, I clearly see it's using default one. So a more precise assert would have made it fail as expected.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: ACK

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
```diff
@@ -210,7 +200,7 @@ public InputStream getInputStream() throws IOException {
 
 	private ProjectBuildingRequest newProjectBuildingRequest() {
 		ProjectBuildingRequest request = new DefaultProjectBuildingRequest();
-		request.setSystemProperties(systemProperties);
+		request.setSystemProperties(lemminxMavenPlugin.getSystemProperties());
```
mickaelistria, line 203: It would be better to use `mavenRequest.getSystemProperty()` here, so we only have to care about a single place.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,61 @@
...
+	@Test
+	public void testUserSettingsWithEnvVariables() throws Exception {
+		final File localSettingsFile = new File(getClass().getResource("/settings-local-repo-test.xml").toURI());
+		String localSettingsContent = FileUtils.readFileToString(localSettingsFile, "UTF-8");
+		String localRepoProperty = System.getProperty(MAVEN_LOCAL_REPO_PROPERTY_NAME);
+		Properties properties = System.getProperties();
+		try {
+			properties.remove(MAVEN_LOCAL_REPO_PROPERTY_NAME); // Not to use local repository set from properties
+			
+			// Add a DUMMY environment variable to check later if it's really resolved to a real path
```
mickaelistria, line 38: It's a system property here, but the test is still useful.
As it's hard to change current env from inside Java, it could be interesting to write multiple tests:
* 1 for system property substitution (eg this one)
* another for env variable (here it could be any existing environment variable, let's just avoid $HOME which is used in default)

So it would test things separately and would bring a better coverage.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -235,14 +246,52 @@ private MavenExecutionRequest initMavenRequest(PlexusContainer container, XMLMav
...
+	 * <p>
+	 * {@link Properties#entrySet()} iterator is not thread safe and fails with
+	 * {@link ConcurrentModificationException} if the source properties "is
+	 * structurally modified at any time after the iterator is created". The
+	 * solution is to use thread-safe {@link Properties#stringPropertyNames()}
+	 * enumerate and copy properties.
+	 *
+	 * @see https://bugs.eclipse.org/bugs/show_bug.cgi?id=440696
+	 */
+	public Properties getSystemProperties() {
```
mickaelistria, line 273: I don't think this would have to be public, nor that a field is required, if we use `mavenRequest.getProperties()` later. This could be a `private static` method.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,61 @@
...
+			properties.put(TEST_DUMMY_PROPERTY_NAME, userHomeProperty);
+			System.setProperties(properties);
+			
+			MavenLemminxExtension plugin = new MavenLemminxExtension();
+			plugin.settings.setGlobalSettings(null);
+			plugin.settings.setUserSettings(localSettingsContent);
+			
+			File localRepositoryPath = plugin.getMavenSession().getRequest().getLocalRepositoryPath();
+			assertNotNull(localRepositoryPath);
+			assertTrue(localRepositoryPath.getAbsolutePath().equals(userHomeProperty + TEST_LOCAL_REPO_PATH));
```
mickaelistria, line 50:
```suggestion
assertTrue(userHomeProperty + TEST_LOCAL_REPO_PATH), localRepositoryPath.getAbsolutePath());
```

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,61 @@
...
+			
+			// Add a DUMMY environment variable to check later if it's really resolved to a real path
+			String userHomeProperty = properties.getProperty(USER_HOME_PROPERTY_NAME);
+			assertNotNull(userHomeProperty);
+			properties.put(TEST_DUMMY_PROPERTY_NAME, userHomeProperty);
+			System.setProperties(properties);
+			
+			MavenLemminxExtension plugin = new MavenLemminxExtension();
+			plugin.settings.setGlobalSettings(null);
+			plugin.settings.setUserSettings(localSettingsContent);
```
mickaelistria, line 46: A file is expected for settings (doc is missing), so it should be
```suggestion
			plugin.settings.setUserSettings(localSettingsFile.getAbsolutePath());
```

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: Yes, you're right
Removed

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java
vrubezhny: ACK

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: ACK

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: I suppose, you mean `assertEquals`?

mickaelistria: yes, edited.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: Hm. It doesn't work like this....
```
org.opentest4j.AssertionFailedError: expected: </home/user/.m2/bug383repository> but was: </home/user/projects/lemminx/lemminx-maven/lemminx-maven/${env.DUMMY_USER_HOME}/.m2/bug383repository>
	at org.eclipse.lemminx.extensions.maven.InitMavenRequestTest.testUserSettingsWithEnvVariables(InitMavenRequestTest.java:51)
```

mickaelistria: That's the same issue I see, and the same one on CI. So overall it looks like it's a good thing you can reproduce it as well.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: There's not many env. variables we  can use... If not creating/setting the new ones,  at the first look we can use `MAVEN_PROJECTBASEDIR` which is set to the path to the current project (lemminx-maven in this case), or something like `PWD`, which is not a good one, imho, as it will make the test to create a `.m2/...` repository in current directory (which is a random thing that depends only on user that starts the script). I wouldn't add the cleaning/removal procedure for the repository here as, in theory, doing so we can remove the m2 repository that is normally used by other tests.

mickaelistria: Does the test create the repository? I was expecting that repository would be created lazily upon request, and here there is no request made to the repository. So if repo is not created, any env var can fit for testing (ie `${HOSTNAME}/.m2/bug383repository` and then comparing to `System.getEnv("HOSTNAME") + "/.m2/bug383repository"`

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: Actually, I see that you're right and the absolute config path is to be used here.
Now I have two problems.... 1) The system properties aren't get resolved in paths (that DUMMY property I have added) - trying to find out why.
2) The second test case added for env. variable added (with the value of maven project location) goes successfully (the var is resolved), but fails on that repository is not accessible - I don;t understand why (is that directory is kinda read only for test or there was some problem on create - not exceptions logged though)

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: Not sure if the repo is really created. For sure the test was creating it when I was invoking `initialize()` method of the Extension.
Now (as I said in other comment) - it is not created (and is not accessible as the result) But ptobably it's even better as we can use almost any env. variable to test it (and I just don't need to call `assertTrue(localRepositoryPath.canRead());` at the end of test case)

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: Nevermind. I have found an issue in my code, so the system/env. properties was not taking into account on interpolate() calls. Now fixed.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: The repositories aren't really created, probably, until the Maven Request is executed. We're only create the requests during the test. So no need to check the resultant  local repository for being accessible.

mickaelistria:
Looks like this is now working, however another test is failing in a topic that may be related.
The current test only check expansion of system properties actually, so it would be good to add a test that uses $HOME env var (as we can't easily use another one).

lemminx-maven/src/test/resources/settings-local-repo-test-env-variable.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0                       http://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <localRepository>${env.MAVEN_PROJECTBASEDIR}/.m2/bug383repository</localRepository>
```
mickaelistria, line 3: This would rather be something like `${env.HOSTNAME}/.m2/bug383repository`

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+	
+	@Test
+	public void testUserSettingsWithEnvVariables() throws Exception {
+		final File localSettingsFile = new File(getClass().getResource("/settings-local-repo-test-env-variable.xml").toURI());
+		String localRepoProperty = System.getProperty(MAVEN_LOCAL_REPO_PROPERTY_NAME);
+		Properties properties = System.getProperties();
+		try {
+			properties.remove(MAVEN_LOCAL_REPO_PROPERTY_NAME); // Not to use local repository set from properties
+			System.setProperties(properties);
```
mickaelistria, line 70: No System property has to/should be set if you want to test env-vars.

mickaelistria (changes requested):

vrubezhny:
> Looks like this is now working, however another test is failing in a topic that may be related. The current test only check expansion of system properties actually, so it would be good to add a test that uses $HOME env var (as we can't easily use another one).

I'l look at the failing test...

But I haven't got the idea why we need this new test with `${HOME}` variable... I just pushed the commit in which I got rid of using any  of `${HOME}` or `${user.home}` because they are the default fallback values. The repository isn't really created  during the test... What should we test with `${HOME}` now?

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/InitMavenRequestTest.java
vrubezhny: If you mean `MAVEN_LOCAL_REPO_PROPERTY_NAME`:
```
private static final String MAVEN_LOCAL_REPO_PROPERTY_NAME = "maven.repo.local";
```
then we have to remove this system property, because otherwise it will be used to create a Maven Local Repo overridinf the values provided by configuration files.

mickaelistria: Ah yes, my bad.

mickaelistria:
> What should we test with ${HOME} now?

The idea is to test a real environment variable, which is also available when running test from Eclipse IDE.

lemminx-maven/src/test/resources/settings-local-repo-test-env-variable.xml
vrubezhny: If we do so then we'll be testing the same `${HOSTNAME}` value to create a local repo path in both - system property  and environment variable tests (I've already used it it system property test). But if it's OK I can change it to be `/${HOSTNAME}` for both test cases.

vrubezhny:
> > What should we test with ${HOME} now?
> 
> The idea is to test a real environment variable, which is also available when running test from Eclipse IDE.

OK

mickaelistria:
Test and code seems correct. Please check why SimpleModelTest.testSystemPathDiagnosticBug is now failing.

mickaelistria:
Thanks!

PR #250, opened 2022-01-19
Remove dependencies on indexer

mickaelistria:


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/ArtifactWithDescription.java
```diff
@@ -0,0 +1,20 @@
+package org.eclipse.lemminx.extensions.maven.participants;
```
akurtakov, line 1: Missing license header.

PR #251, opened 2022-01-19
Switch to use OkHttpClient instead of Unirest-Java

vrubezhny:

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -125,11 +125,6 @@
 			<artifactId>wagon-file</artifactId>
 			<version>3.4.3</version>
 		</dependency>
-		<dependency>
-            <groupId>com.konghq</groupId>
-            <artifactId>unirest-java</artifactId>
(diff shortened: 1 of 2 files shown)


mickaelistria:
I merged a patch removing usage of indexer class, it basically replace usage of ArtifactInfo by org.apache.maven.aether.Artifact. Please try rebasing, hopefully merge is simple.
Also, please remove the various System.err.println. If they're often useful, you may want to use `LOGGER.finest(...)` to log those, and configure the tests so the the "finest" level is enabled.

vrubezhny:
@mickaelistria Done.

fbricon:
Why not use java 11's built-in http client?

mickaelistria:
> Why not use java 11's built-in http client?

That would be a good improvement; but as okhttp3 is there and loaded anyway when trying to fetch artifacts, I don't think there would be much immediate win. But indeed, we'll need to get rid of okhttp3 if Maven itself stops using it.

Issue #253, opened 2022-01-21
Plugins cannot be opened by Ctrl+<Click> if version is not explicitly specified


HannesWell:
If a plug-in is specified in the `pluginManagment` or `plugins` section its pom.xml can only be opened by `CTRL+<Click>` if it's version is explicitly specified in place.
It cannot be opened if its version is either a property or inherited from another `pluginManagment`.

It can also not be opened if its `groupId` is `org.apache.maven.plugins` but its `groupId` element is omitted  which is legal for plug-ins from that group.

Example `pom.xml`:
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>foo.bar</groupId>
	<artifactId>test-click-open</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<properties>
		<compiler.plugin.version>3.9.0</compiler.plugin.version>
		<antrun.plugin.version>3.0.0</antrun.plugin.version>
	</properties>
	<build>
		<pluginManagement>
			<plugins>
				<plugin>
					<groupId>org.apache.maven.plugins</groupId>
					<artifactId>maven-jar-plugin</artifactId>
					<version>3.2.2</version>
				</plugin>
				<plugin>
					<groupId>org.apache.maven.plugins</groupId>
					<artifactId>maven-antrun-plugin</artifactId>
					<version>${antrun.plugin.version}</version>
				</plugin>
				<plugin>
					<artifactId>maven-assembly-plugin</artifactId>
					<version>3.3.0</version>
				</plugin>
			</plugins>
		</pluginManagement>
		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-clean-plugin</artifactId>
				<version>3.1.0</version>
			</plugin>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>${compiler.plugin.version}</version>
			</plugin>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-antrun-plugin</artifactId>
			</plugin>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
[...]

PR #262, opened 2022-02-07
Plugins cannot be opened by Ctrl+<Click> if version is not explicitly…

vrubezhny:
… specified #253


mickaelistria:
Great, thansk!

Issue #259, opened 2022-02-02
Hovering on a dependency doesn't show info is version is a property


mickaelistria:
Hovering on artifactId of eg
```xml
		<dependency>
			<groupId>org.apache.maven</groupId>
			<artifactId>maven-embedder</artifactId>
			<version>${maven.version}</version>
			<scope>compile</scope>
		</dependency>
```
in a valid pom file gives no result, also the `${maven.version}` is properly resolved and project builds properly (resolves the dependency).

PR #270, opened 2022-02-16
Hovering on a dependency doesn't show info if version is a property #259

vrubezhny:
Issue: #259 


mickaelistria:
StringBuilder.isEmpty was introduced in Java 15. Please fix the code to use some construct that's available in Java 11.

mickaelistria:
>  It's not reported in Eclipse even though Java 11 compatibility is se

You need to have an actual Java 11 configured in Eclipse IDE for that. Then JDT will use the classes from the Java 11 installation and see this method is missing.

PR #272, opened 2022-02-23
Maven Projects are built and cached more than once  when relative pat…

vrubezhny:
…hs are used

It's possible that the same maven project is built and cached two or more times when its URI  calculated relatively to its child proect.
For example, for the following URIs:

	file:/home/user/projects/maven-hierarchy-test/child/grandchild/../../pom.xml
	file:/home/user/projects/maven-hierarchy-test/pom.xml

we'll have two copies of the same maven project built and saved into the Maven Project Cache


mickaelistria:
That's a great finding and patch!

Issue #276, opened 2022-02-25
Existing elements of an existing dependency got broken when inserting `artifactId` from CA  


vrubezhny:
Steps to reproduce:
- Consider editing Maven Project like the following:

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <artifactId>foo-child</artifactId>
  <packaging>pom</packaging>
  <name>FOO Child</name>
  <dependencies>
	<dependency>
		<artifactId>org.ecl|</artifactId>
	</dependency>
  </dependencies>
</project>
``` 
- Place the cursor in text instead of "|" character and press Ctrl-Space to invoke the content assistant
- Continue typing characters, for example `ipse.lem` in order to make `org.eclipse.lemminx`  artifact to be the first selected proposal, then insert the selected proposal
- See the proposal inserted breaking the existing code:
```
	<dependency>
		<groupId>org.eclipse.lemminx</groupId>
		<artifactId>org.eclipse.lemminx</art
		<version>0.19.1</version>ifactId>
	</dependency>
```

vrubezhny:
The main fix is made in LSP4E: https://github.com/eclipse/lsp4e/issues/89 
The PR #285 contains additional fix for `<version> element in CA
The PR #286 contains new JUnit  test cases made for the issue (depends on lemminx project PR - https://github.com/eclipse/lemminx/pull/1186)

PR #277, opened 2022-03-03
Existing elements of an artifact dependency got broken when …

vrubezhny:
…inserting artifactId from CA #276

Issue: #276

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java
@@ -476,13 +476,27 @@ public class MavenCompletionParticipant extends CompletionParticipantAdapter {
 				if (insertVersion) {
 					Position insertionPosition;
 					try {
-						insertionPosition = request.getXMLDocument()
+						Position position = request.getXMLDocument()
 								.positionAt(request.getParentElement().getEndTagCloseOffset() + 1);
+						insertionPosition = new Position(position.getLine() + 1, 0);
+						int insertsoinOffset = request.getXMLDocument().offsetAt(insertionPosition);
+						int end = request.getXMLDocument().getEnd();
+						
+						// In case of the current line is the last document line and
+						// it doesn't have an EOL character typed - we need to insert 
+						// such am EOL character in order to keep text formatting
+						boolean insertEOL = insertsoinOffset >= end;
+						if (insertEOL) {
+							String  lineDelimiter = request.getXMLDocument().lineDelimiter(position.getLine());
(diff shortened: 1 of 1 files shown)


mickaelistria:
Do you think you could write a test case that reproduce the initial issue, so we are sure there will be no regression for this case?

vrubezhny:
It's a bit problematic... Yes I'm working on a JUnit test that is checking the correctness of `CompletionItem` generated on Lemminx-Maven side, but it's not the final process destination - those `CompletionItem`-s are to be transferred to the client and converted for example into `LSIncompleteCompletionProposal` which is also processed (f.i.,  according to the number of characters typed in from the moment of the content assist is invoked) and then proposal apply which could fail (f.i., with a `  BadLocationException`). So, having such a test still doesn't add any guaranty that the issue is really resolved.

mickaelistria:
I'm confused. What the issue here: is it lemminx-maven sending incorrect CompletionItems (and this patches fixes it), or LSP4E not applying them correctly, or both?

vrubezhny:
A problem can happen in 'extreme' situations - like insertion at line/document ending positions. In Lemminx you never know if a document line really has an EOL character   (https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocument.java#L89) - `TextDocument` always returns at least non-null default value. This creates a possibility for  `BadLocationException`-s to happen when inserting a text at a positions like "at next line".

This exact issue, is actually probably a LSP4E issue which incorrectly updates positions for a proposal that consists of multiple additional `TextEdit`-s (like we have for the current case - we need to insert `<group/> element before the main `TextEdit` and a `<version/>` element after). I'm sure `LSIncompleteCompletionProposal.apply()` tries to update positions before the insertions according to the number of characters typed after the opening of CA, but it does it incorrectly for some reason.
This made me think to change the `TextEdits` calculation for the case in lemminx-maven - now I'm trying not to use the line end position for the `<version/>` element `TextEdit` trying to always make this insertion to 'the beginning of the next line'. It worked because not an offset, but positions (line,character) are transferred between Lemminx-Maven and LSP3E.
But in this case I need to know if the last  line really has an EOL char inserted - if it doesn't I must explicitly add such am EOL char, otherwise the insertion will occur on not existing position on LSP4E side, so a "properly" created proposal will not be really inserted.

mickaelistria:
> This exact issue, is actually probably a LSP4E issu

OK, so please report it to LSP4E. We won't add hacks to lemminx-maven for issues that are in LSP4E, instead, we'll fix LSP4E.

vrubezhny:
Closing this PR as the main fix is made in LSP4E: https://github.com/eclipse/lsp4e/issues/89 as well as additional fixed are provided in #285 and `lemminx` project - https://github.com/eclipse/lemminx/pull/1186

PR #285, opened 2022-03-11
Existing elements of an artifact dependency got broken when … #285

vrubezhny:
... from CA #276
    
A complimentary fix for `<version/>` element inserted by <artidactId/> CA proposal
    
Issue: #276


mickaelistria:
Is there a way to get this patch using some released version of LemMinX?

vrubezhny:
I can break the patch into two parts - the patch itself and JUnit test part, then the fix can be applied immediately, while the PR with JUnit test can wait for lemminx to be released

vrubezhny:
JUnit test cases for the issue are moved to PR #286

vrubezhny:
@mickaelistria  I did it before moving changes to `lemminx`... 

To do that I need to drag 4 methods from `org.eclipse.lemminx.XMLAssert` class to `MavenCompletionParticipantTest`. and directly import few `org.lsp4j.*` classes.. 
Not the best solution, IMHO, unless it's a temporary workaround...
But, yes. I can do it this way again.

Issue #238, opened 2021-12-17
Add Managed Version Info on hover


mickaelistria:
See https://github.com/eclipse-m2e/m2e-core/issues/437

mickaelistria:
Fixed with https://github.com/eclipse/lemminx-maven/commit/025bc243df52b2b4145c5aaa594bfd0c53781d75

PR #298, opened 2022-08-26
Add Managed Version Info on hover #238

vrubezhny:

### DIFF
--- a/lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
+++ b/lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lemminx.extensions.maven.participants.hover;
+
+import static org.eclipse.lemminx.extensions.maven.utils.MavenLemminxTestsUtils.createDOMDocument;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.extensions.maven.NoMavenCentralExtension;
+import org.eclipse.lemminx.services.XMLLanguageService;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.settings.XMLHoverSettings;
+import org.eclipse.lsp4j.Hover;
+import org.eclipse.lsp4j.HoverCapabilities;
+import org.eclipse.lsp4j.MarkupContent;
+import org.eclipse.lsp4j.MarkupKind;
+import org.eclipse.lsp4j.Position;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(NoMavenCentralExtension.class)
+public class ManagedVersionHoverTest {
+
+	private XMLLanguageService languageService;
+
+	@BeforeEach
+	public void setUp() throws IOException {
+		languageService = new XMLLanguageService();
+	}
+
+	@AfterEach
+	public void tearDown() throws InterruptedException, ExecutionException {
+		languageService.dispose();
+		languageService = null;
+	}
+
+	@Test
+	@Timeout(90000)
+	public void testManagedVersionHoverInDependencyChild() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-dependencyManagement-child.xml", languageService);
+		Hover hover = languageService.doHover(document, new Position(15, 25), createSharedSettings());
+		String value = ((MarkupContent) hover.getContents().getRight()).getValue();
+		System.out.println("testManagedVersionHoverInDependencyChild: [" + value + "]");
+		assertNotNull(value);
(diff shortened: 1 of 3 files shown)


angelozerr:
Just a simple comment when you create a PR. I think it should be nice to add a little screenshot with th eresult of PR (or a little demo). It will be easier to understand what the PR provides and for history it is nice too.

We do that for LemMinX and I find it is nice for the reviewer.

It is just a suggestion that I do.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
```diff
@@ -0,0 +1,103 @@
...
+	public void testManagedVersionHoverInPluginParent() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-pluginManagement-parent.xml", languageService);
+		Hover hover = languageService.doHover(document, new Position(14, 33), new SharedSettings());
+		String value = ((MarkupContent) hover.getContents().getRight()).getValue();
+		System.out.println("testManagedVersionHoverInPluginParent: [" + value + "]");
+		assertNotNull(value);
+		assertTrue(value.contains("The managed version is"));
+		assertTrue(value.contains("2.22.2"));
+		assertTrue(value.contains("The artifact is managed in"));
+		assertTrue(value.contains("org.apache.maven.plugins:maven-surefire-plugin:2.22.2"));
```
mickaelistria, line 101: I don't think this should be added in when the version is the one that is defined in the pom. Managed version are only useful when no version is explicitly set (usually in the "child pom")

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
```diff
@@ -0,0 +1,103 @@
...
+	public void testManagedVersionHoverInPluginChild() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-pluginManagement-child.xml", languageService);
+		Hover hover = languageService.doHover(document, new Position(18, 33), new SharedSettings());
+		String value = ((MarkupContent) hover.getContents().getRight()).getValue();
+		System.out.println("testManagedVersionHoverInPluginChild: [" + value + "]");
+		assertNotNull(value);
+		assertTrue(value.contains("The managed version is"));
+		assertTrue(value.contains("2.22.2"));
+		assertTrue(value.contains("The artifact is managed in"));
+		assertTrue(value.contains("org.apache.maven.plugins:maven-surefire-plugin:2.22.2"));
```
mickaelistria, line 87: I think what is expected here is more the reference to the parent pom that adds the version management (here pluginManagement-parent.xml)

mickaelistria (changes requested):
Unless I missed something, I believe some extra informations are confusing and should either be fixed or removed.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
vrubezhny: But we still have to show the version and the location links on a pom where the version is clearly defined.

That was actually a problem as the method I've used for finding version in such case was returning a wrong version and link. So, I'd prefer this test case to also exist for parent pom.

mickaelistria: > But we still have to show the version and the location links on a pom where the version is clearly defined.

we already have that I believe. Ctrl+Click can be used to get the link to the artifact.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
```diff
@@ -0,0 +1,100 @@
...
+	public void testManagedVersionHoverInDependencyChiild() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-dependencyManagement-child.xml", languageService);
+		Hover hover = languageService.doHover(document, new Position(15, 25), new SharedSettings());
+		String value = ((MarkupContent) hover.getContents().getRight()).getValue();
+		System.out.println("testManagedVersionHoverInDependencyChiild: [" + value + "]");
+		assertNotNull(value);
+		assertTrue(value.contains("The managed version is"));
+		assertTrue(value.contains("2.22.2"));
+		assertTrue(value.contains("The artifact is managed in"));
+		assertTrue(value.contains("org.apache.maven.plugins:maven-surefire-plugin:2.22.2"));
```
mickaelistria, line 60: That's not correct. The artifact version is managed in pom-dependencyManagement-parent.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
```diff
@@ -0,0 +1,100 @@
...
+	@Test
+	@Timeout(90000)
+	public void testManagedVersionHoverInDependencyParent() throws IOException, URISyntaxException {
+		DOMDocument document = createDOMDocument("/pom-dependencyManagement-parent.xml", languageService);
+		Hover hover = languageService.doHover(document, new Position(13, 29), new SharedSettings());
+		String value = ((MarkupContent) hover.getContents().getRight()).getValue();
+		System.out.println("testManagedVersionHoverInDependencyParent: [" + value + "]");
+		assertNotNull(value);
+		assertFalse(value.contains("The managed version is"));
+		assertFalse(value.contains("The artifact is managed in"));
```
mickaelistria, line 72: good.

mickaelistria (changes requested):

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
vrubezhny: Now a version from the parent pom and location of parent pom where the version is defined is shown in Hover.

If no version is specified in project it  shows a version and location of pom for an according artifact resolved in maven project.

lemminx-maven/src/test/java/org/eclipse/lemminx/extensions/maven/participants/hover/ManagedVersionHoverTest.java
vrubezhny: Fixed

mickaelistria:
Thanks!

pbodnar:
Thank you!

It would be great to also jump directly to the line where the given artifact version is defined and not to just open the linked pom.xml. A new feature request targeted on LemMinX, on m2e, or maybe both?

PR #302, opened 2022-10-06
Add dependabot config

akurtakov:


akurtakov:
@HannesWell how to make licensecheck run in subdir?

HannesWell:
> I have reduced this one to dependabot only so there is at least some automation enabled while I look at it. @HannesWell I would appreciate if you spend the time to provide PR that enables licensecheck on this repo.

I think this can addressed if I add a parameter to force another root pom file to the license-check build that is then used as value for the `-f` option and defaults to `pom.xml`. For this repo the value could then be set to `lemminx-maven`.

I will test and implement that (probably) this evening and will let you know if it's done.

mickaelistria:
@HannesWell I think it's not worth bothering too much with it. https://github.com/eclipse/lemminx-maven/issues/311 seems like a better approach.

PR #312, opened 2022-10-07
Move code to root folder

mickaelistria:


HannesWell:
Don't you think it is an option to just add a 'root' pom.xml to the repo root and make lemminx-maven the only module?
Having Maven-Projects directly in the git-repo root is not common and makes future additions of modules difficult respectivly impossible without moving the code back.

akurtakov:
> Don't you think it is an option to just add a 'root' pom.xml to the repo root and make lemminx-maven the only module? Having Maven-Projects directly in the git-repo root is not common and makes future additions of modules difficult respectivly impossible without moving the code back.

That's fair point.

mickaelistria:
OK

PR #314, opened 2022-10-07
Add licensecheck

mickaelistria:


HannesWell:
> /request-license-review

IIRC that only works if the workflow is on the master branch. But you can request licenses even if the PR is closed. Then the master branch is checked.

HannesWell:
It kicks now, I'm just not a committer and therefore the request is rejected.
For for the latested commit on the master the checked succeeds.

PR #354, opened 2023-01-19
infinite loop when hover

zollum:
infinite loop when hover dependency managed in grandparent's pom.


vrubezhny (approved):
Looks good to me.
Thanks!

Issue #360, opened 2023-02-09
java.lang.NoSuchMethodError: [...] RemoteRepository$Builder.setBlocked(boolean) on completion


mfussenegger:
When triggering completion I'm getting the following error:

```
java.util.concurrent.CompletionException: java.lang.NoSuchMethodError: 'org.eclipse.aether.repository.RemoteRepository$Builder org.eclipse.aether.repository.RemoteRepository$Builder.setBlocked(boolean)'
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:649)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:483)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.NoSuchMethodError: 'org.eclipse.aether.repository.RemoteRepository$Builder org.eclipse.aether.repository.RemoteRepository$Builder.setBlocked(boolean)'
	at org.apache.maven.RepositoryUtils.toRepo(RepositoryUtils.java:213)
	at org.apache.maven.bridge.MavenRepositorySystem.getProxy(MavenRepositorySystem.java:279)
	at org.apache.maven.bridge.MavenRepositorySystem.injectProxy(MavenRepositorySystem.java:310)
	at org.apache.maven.internal.aether.DefaultRepositorySystemSessionFactory.newRepositorySession(DefaultRepositorySystemSessionFactory.java:221)
	at org.eclipse.lemminx.extensions.maven.MavenLemminxExtension.initialize(MavenLemminxExtension.java:174)
	at org.eclipse.lemminx.extensions.maven.MavenLemminxExtension.getLocalRepositorySearcher(MavenLemminxExtension.java:400)
[...]

PR #361, opened 2023-02-10
Exclude aether-api and aether-util for zip-with-dependencies package

mfussenegger:
I'm not sure why, but having these jars in the classpath lead to
`NoSuchMethodError` exceptions when triggering completion.

Fixes https://github.com/eclipse/lemminx-maven/issues/360

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -131,14 +131,22 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
-			<groupId>org.eclipse.aether</groupId>
-			<artifactId>aether-connector-basic</artifactId>
-			<version>1.1.0</version>
-		</dependency>
-		<dependency>
 			<groupId>io.takari.aether</groupId>
 			<artifactId>aether-connector-okhttp</artifactId>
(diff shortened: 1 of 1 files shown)


lemminx-maven/pom.xml
```diff
@@ -130,15 +130,20 @@
...
-		<dependency>
-			<groupId>org.eclipse.aether</groupId>
-			<artifactId>aether-connector-basic</artifactId>
-			<version>1.1.0</version>
-		</dependency>
 		<dependency>
 			<groupId>io.takari.aether</groupId>
 			<artifactId>aether-connector-okhttp</artifactId>
 			<version>0.17.8</version>
+			<exclusions>
```
mickaelistria, line 137: Can you please add a comment mentioning the issue; so we easily know later why this shouldn't be modified blindly?

mickaelistria (changes requested):
I think it's ok to merge it. Just please add a note about "why" in the exclusion.

lemminx-maven/pom.xml
mfussenegger: Done

mickaelistria:
Thank you!

PR #362, opened 2023-02-15
Add code action participant

vrubezhny:

PR #365, opened 2023-02-18
Add Code Actions for Parent Version/Group ID removal 

vrubezhny:
This PR adds Maven project validation for groupId/version element values duplicating the parent groupId/version element values
Also this adds Code Actions for the removal of the duplicating groupId/version elements and according JUnit test cases

Depends on #362

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
@@ -64,6 +64,7 @@ import org.codehaus.plexus.component.repository.exception.ComponentLookupExcepti
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.WorkspaceReader;
 import org.eclipse.lemminx.dom.DOMDocument;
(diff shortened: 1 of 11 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -155,6 +158,8 @@ public void start(InitializeParams params, XMLExtensionsRegistry registry) {
 			registry.registerDiagnosticsParticipant(diagnosticParticipant);
 			hoverParticipant = new MavenHoverParticipant(this);
 			registry.registerHoverParticipant(hoverParticipant);
+			codeActionParticipant = new MavenCodeActionParticipant();
+			registry.registerCodeActionParticipant(codeActionParticipant);
```
mickaelistria, line 162: As mentioned on the other patch, it should be preferred to register each participant individually instead of grouping them.

vrubezhny: Yes. this is subject to change

mickaelistria:
This looks OK. Anything left to do before this PR can be turned into "ready for review"?

vrubezhny:
License vetting status check job didn't run yet. Should be OK, but it's better to have it succeeded before merge

Issue #267, opened 2022-02-10
Completion should also propose other packaging types available in Build extensions


mickaelistria:
Similarly to https://github.com/eclipse-m2e/m2e-core/commit/bf03a75b8cdd2d8494408e0ae42512bd1bd7bbf8 , completion for packaging type should lookup in build extensions to resolve extra packaging types.

PR #386, opened 2023-04-12
Completion should also propose other packaging types available in Bui…

vrubezhny:
…ld extensions #267

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/DOMConstants.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/DOMConstants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Red Hat Inc. and others.
+ * Copyright (c) 2021-2023 Red Hat Inc. and others.
  * This program and the accompanying materials are made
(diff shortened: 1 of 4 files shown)


mickaelistria:
One great test you could add would be to add a pom that defines Tycho as build extension plugin and check whether the `eclipse-plugin` type is available in completion.
I don't think it's so interesting to list other non-native packaging types if those can be inferred dynamically.

vrubezhny:
Yes, I'm thinking on the test. Problem could be that MavenProject is only partially setup (because of `org.apache.maven.project.DefaultProjectBuilder` that isn't tolerant when parsing an incomplete document, thus not returning any result in case of packaging element is empty (throwing an `IllegalArgumentException` immediately after `ModelBuildingException` related to the empty packaging occurs - that's why I disabled the dependencies resolution on building the maven project). 
So a better test is needed for non-native packaging types.

vrubezhny:
`MavenCompletionParticipantTest.testPackagingyCompletion` is added with the detection of `eclipse-plugin` packaging type defined in `org.eclipse.tycho:tycho-maven-plugin:2.7.5`

mickaelistria (approved):
Great addition!

PR #395, opened 2023-04-26
Fix Workspace POM files processing ordering

vrubezhny:
Do not add any sorting forPOM files during 'MavenLemminxWorkspaceReader.addToWorkspace()' work. The files are to be processed in the same order as they specified in method argument. Any sorting is to be done before calling the method.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxWorkspaceReader.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxWorkspaceReader.java
@@ -16,11 +16,13 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
@@ -155,7 +157,11 @@ public class MavenLemminxWorkspaceReader implements WorkspaceReader {
 	private final WorkspaceRepository repository;
 	private final MavenLemminxExtension plugin;
 	
-	private SortedSet<File> toProcess = Collections.synchronizedSortedSet(new TreeSet<>(Comparator.comparingInt(file -> file.getAbsolutePath().length())));
+	// Don't add any sorting here, the files are to be processed exactly in the order they added.
+	// Any sorting is to be done before adding to this set
+	//
+	private Set<File> toProcess = Collections.synchronizedSet(new LinkedHashSet<File>());
+	
 	private ThreadLocal<Boolean> skipFlushBeforeResult = new ThreadLocal<>();
 	private final PriorityBlockingQueue</*ResolveArtifactsAndPopulateWorkspaceRunnable*/Runnable> runnables = new PriorityBlockingQueue<>(1, DEEPEST_FIRST);
 	private final ExecutorService executor = new ThreadPoolExecutor(1, 1, 0, TimeUnit.MILLISECONDS, runnables);
@@ -278,6 +284,7 @@ public class MavenLemminxWorkspaceReader implements WorkspaceReader {
 
 	/**
 	 * Parses and adds a document for a given URI into the projects cache
+	 * Any sorting is to be done before the method is invoked.
 	 * 
 	 * @param uri An URI of a document to add
 	 * @param documents documents to add


mickaelistria:
The sorting was a heuristic to improve performance as usually, the shorter the path is, and the more likely the pom is a parent of another.
So if you have a structure

* parent
  ** child1
  ** child2

this sorting will ensure you build first parent, then child1 then child2; which would usually be the right order. If you send child1, child2 then parent, then projects will need to be rebuilt according to the local copy of the parent, so processing them would result in child1 being built, then child2 being build, then parent being built, then child1 being rebuilt then child2 being rebuilt.
Basically, this optimization is intended to make processing the projects O(n) instead of O(n^2) in usual cases.

mickaelistria:
> Unfortunately sorting based on pom file path length doesn't work good enough.

Do you have an example of a case where this sort would be bad?

> This PR makes the ResolveArtifactsAndPopulateWorkspaceRunnable runnables to be executed in the same order as they arrived from didChangeWorkspaceFolders's sorting

How does it improve things? As mentioned in my previous comment, I see it can easily turn some O(n) cases into O(n²) , so we need to make sure there is serious benefit to balance that cost.

vrubezhny:
> > Unfortunately sorting based on pom file path length doesn't work good enough.
> 
> Do you have an example of a case where this sort would be bad?
> 
https://github.com/eclipse/lemminx-maven/tree/master/lemminx-maven/src/test/resources/issue-345


> > This PR makes the ResolveArtifactsAndPopulateWorkspaceRunnable runnables to be executed in the same order as they arrived from didChangeWorkspaceFolders's sorting
> 
> How does it improve things? As mentioned in my previous comment, I see it can easily turn some O(n) cases into O(n²) , so we need to make sure there is serious benefit to balance that cost.

It depends on number of projects to be processed during the workspace loading. 
I'm sorting project URIs before their "Resolve" runnables are added to Workspace Reader's queue. So now the first added URI points to one of the "parent" projects. 
Problem is that while the Workspace Reader resolves the first project, the rest of the runnables added to the queue got re-ordered (on file path length based sorting) - so, at the time the first parent URI is processed, the next URI can be a child of some other parent (because the queue is re-ordered). which can lead in failure of creating a MavenProject for that child and leaved it not cached (Diagnostics/CA/Hovers etc. - not working)

mickaelistria:
I think I understand the issue. And I agree the order of initial request is worth being kept and honored. However, I suggest the following heuristic to make the usual case happy: if a project is requested and is in a parent directory of another requested one, or if it's in a sibling directory but whose name contains "parent", the we try to insert it in the list before.
What do you think?

mickaelistria:
> Maven doesn't operate with file/directory names - the names are meaningless.

Sure, but most users do put meaning in their filesystem structure.

> When we read a user's workspace

I agree it makes sense, but the startup if a workspace is not the only reason why the pom gets refreshed. didSave also plays here.
Imagine a case with a long list of pom files in the workspace and the 1st one is a parent pom in a parent directory, and while other poms are being processed in background, user modifies the parent pom and saves. `didSave` is sent. Then for best result, we do want this parent to get processed ASAP, before it's probable children. However, if user did call `didSave` but other poms are being processed from initialization, we can imagine that this pom has high ("user") priority because it's being worked on, so we should try to prioritize its processing over the pending ones. I don't think this current change allows that (nor does the current state).
Trying to optimize the order here is an important criterion of perceived performance, I think we should maybe have dedciated tests to cover the possible cases we want to optimize before changing it.

vrubezhny:
On `didSave` we're just re-reading the settings and then `stop()` and re-`start()` the extension using the `initialWorkspaceFolders` that were set on the first start (so next `initialize()` invocation will peek and parse them again (and the only newly saved pom.xml file is probably going to be resolved again because of its new document version)...

I don't think its going to happen on start (during the first workspace POM URI's resolve pass), but if it is - we still need to parse them all one by one (some of them will be already cached, so it will take less time to reach the document just saved)

mickaelistria:
> We should build a chain like "parent_of_saved_document -> its_yet_another_parent .-> .. -> saved_document", then check if every parents are built and cached (process them if they are not) and only then build that "saved_document" - only in this case we'll be successful on building it and thus we can insert such a chain into the beginning of the queue - so these added POM files to be processed as soon as possible.

Sounds OK (until we find something better).
But this discussion highlights that the is no 1 standard data structure (sorted set, queue, priorityqueue...) that can cover all the possible cases we want to optimize, so we do probably need to encapsulate better the list of projects to process (most likely in some new PendingProjects class) and extract a method `getNextProject()` from it where we can implement such smarter logic. That would allow to dynamically reevaluate whether there is now 1 more interesting project to process.

vrubezhny:
@mickaelistria thanks. I agree with your proposal.
So, I'm going to merge the PR now as a temporary solution.

Issue #358, opened 2023-02-01
Cancel any operation as soon as possible.


angelozerr:
Once LSP4E will support cancel support https://github.com/eclipse/lsp4e/pull/431, it should be really nice to support cancel support on lemminx-maven side.

LemMinx support cancel support which stop any operation like XML completion, diagnostics as soon as possible. As vscode provides cancel support, we have a very good performance.

It misses only hover (see https://github.com/eclipse/lemminx/issues/1474)

It is very easy to support cancel support. For instance indefinition you receive the cancel checker https://github.com/eclipse/lemminx-maven/blob/dd3fe65991ad39cbf45c7f689a0a9e49500a4463/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/definition/MavenDefinitionParticipant.java#L52

and in long process you should call cancelChecker.checkedCanceled(); like in https://github.com/eclipse/lemminx-maven/blob/dd3fe65991ad39cbf45c7f689a0a9e49500a4463/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/definition/MavenDefinitionParticipant.java#L122

angelozerr:
Please see the PR https://github.com/eclipse/lemminx/pull/1502 which provides the cancel  checker in the Hover API participant.

fbricon:
Shouldn't that issue be closed since #398 was merged?

vrubezhny:
@fbricon I don't think so yet. There is still a number of participants to add cancellation checks or re-visited if they can be cancelled as soon as possible.

angelozerr:
@fbricon please note that it will not work on Eclipse IDE side since LSP client doesn't cancel the request.

angelozerr:
@vrubezhny I think we can close this issue?

vrubezhny:
@angelozerr Yes, I believe the most of cancelling support is added, so we can close this issue.  Thank you!

PR #397, opened 2023-04-27
Hover Participant - Cancel any operation as soon as possible.

vrubezhny:
Issue: #358

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/hover/MavenHoverParticipant.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/hover/MavenHoverParticipant.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
 import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 import java.util.logging.Level;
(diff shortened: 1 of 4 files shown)


angelozerr:
I'm glad @vrubezhny that you use this cancel support, but have you checked it is working with LSP4E?

I know that @mickaelistria has done something about this cancel support on LSP client side for hover, but it could be nice to check it is working with LSP4E by debugging the lemminx maven extension.

vrubezhny:
@angelozerr Thanks for the info. 
The JUnit tests pass well, so the cancellation support  doesn't bring any visible problems. but I'll try finding out what @mickaelistria has  done on LSP client side and try debugging the extension.

mickaelistria:
Cancellation is not yet supported in LSP4E, it got better but not yet complete. So at the moment, there is not even 1 client of lemminx-maven that supports cancellation, I suggest we just ignore it for the moment, until we can benefit from it on client side.

angelozerr:
Vscode support cancel support

mickaelistria:
> Vscode support cancel support

As far as I know, lemminx-maven is not yet supported in VSCode, is it?

vrubezhny:
Once the cancellation is not fully supported in LSP4E, I can suggest a simple tests on LemMinX-Maven side, like handling an immediate cancellation that mostly check that the cancelling itself doesn't bring any problems. 
Later we can probably add some more testing.

mickaelistria:
Automated test would be great!

PR #425, opened 2023-06-24
Experiment

cstamas:
Changes:
* Upgrade resolver to 1.9.13 (does not really matter due that below)
* Use no-op locking (so let FS does it's best, will probably explode on Windows)

### DIFF
--- a/lemminx-maven/pom.xml
+++ b/lemminx-maven/pom.xml
@@ -26,7 +26,7 @@
 		<junit-jupiter.version>5.9.3</junit-jupiter.version>
 		<lemminx.version>0.26.1</lemminx.version>
 		<maven.version>3.9.2</maven.version>
-		<maven-resolver.version>1.9.12</maven-resolver.version>
+		<maven-resolver.version>1.9.13</maven-resolver.version>
 	</properties>
 
 	<dependencies>
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
@@ -198,6 +198,7 @@ public class MavenLemminxExtension implements IXMLExtension {
 			this.container = newPlexusContainer();
 			mavenRequest = initMavenRequest(container, settings);
 			DefaultRepositorySystemSessionFactory repositorySessionFactory = container.lookup(DefaultRepositorySystemSessionFactory.class);
+			mavenRequest.getUserProperties().setProperty("aether.syncContext.named.factory", "noop");
 			RepositorySystemSession repositorySystemSession = repositorySessionFactory.newRepositorySession(mavenRequest);
 			MavenExecutionResult mavenResult = new DefaultMavenExecutionResult();
 			// TODO: MavenSession is deprecated. Investigate for alternative


vrubezhny:
@cstamas  are `mavenRequest.getUserProperties().setProperty("aether.syncContext.named.factory", "noop");` and other possible configurations and properties documented somewhere?

cstamas:
Resolver configurations are here: https://maven.apache.org/resolver/configuration.html

I just saw that how you configure resolver session uses mavenRequest, so added it there (as user property, like user would do `-Daether....` in Maven CLI)

cstamas:
Doco for named locks here https://maven.apache.org/resolver/maven-resolver-named-locks/index.html (plus there is redisson and hazelcast in separate module)

cstamas:
But due resolver -> WorkspaceReader -> resolver exec path no other locking than "noop" will work...

vrubezhny:
>  But due resolver -> WorkspaceReader -> resolver exec path no other locking than "noop" will work...

So, do you think  your PR can be applied as a "final"-like solution for the issue?

cstamas:
Without deeper insight, I cannot for sure say "yes" or "no" :smile: I can only hope, that test coverage is (error) detection is correct, and derived from that I could say "probably".

For sure (and for me) the "proper" solution would be to refactor things, and break the resolver -> WSR -> resolver execution path, but again, am not in details of this project.

vrubezhny:
Hm. Locally adding
```
request.getUserProperties().setProperty("aether.syncContext.named.factory", "noop");

```
to  `ProjectBuildingRequest` configuration doesn't change anything. 
I'm still caught with  `IllegalStateException: Could not acquire lock(s)` problem

My changes (including your change to `ProjectBuildingRequest` confg are in https://github.com/eclipse/lemminx-maven/pull/426

vrubezhny:
@cstamas I don't really get what do you mean by `break the resolver -> WSR -> resolver execution path`...

We have a workspace containing user's maven projects. So, when user wants to edit a project (pom.xml file) and opens it in editor (one of the possibilities is use M2E-Core Editor Lemminx):
- Lemminx Language Server stars  and validates the XML file
- Lemminx-Maven extension starts (because it's a Maven project file) and gets the workspace URI on initialization
- Lemminxx-Maven collects the Maven Project URIs in the Workspace, sorts them (just by reading the project's and it's  parent GAVs and putting the parents to be first on the list, then children, then everything not-determined at the end of the list).
- Then this sorted list of Workspace Project URSs come to Workspace Reader which starts the build - projects are build one by one and the results are cached in MavenProjectCache.

So I see the only `WSReader -> resolver` path. Not a `resolver -> WSReader -> resolver`...

Probably, M2E builder can be executing the the same time (as Eclipse builder) .

cstamas:
The `MavenLemminxWorkspaceReader` (WSReader) implements Maven Resolver API WorkspaceReader, that is used within resolver. Hence, `resolver -> MavenLemminxWorkspaceReader` is there implicitly. Hence, if you have `MavenLemminxWorkspaceReader -> resolver`, and you have, implies that you really have `resolver -> MavenLemminxWorkspaceReader -> resolver`.

vrubezhny:
The `WorkspaceReader` API methods implemented in `MavenLemminxWorkspaceReader` do mot build any Maven projects - they use only the information that is already gathered.  The builder is invoked only during initialization, so, imho, it's two separated path to be played at different stages: `WSRreader  (initialization)-> resolver` and `resolver -> WSReader (WorkspaceReader API)` ... But maybe we should ensure that such a separation is really respected

vrubezhny:
@cstamas Thanks for the suggestion to "use no-op locking".

If you don't mind I would incorporate it to PR #426 and close this PR.

vrubezhny:
Closing as the fix is incorporated in #426 

@cstamas Thank you for your  help!

Issue #462, opened 2023-07-27
Cancel Project Validation as soon as Diagnostic Request is cancelled


vrubezhny:
The Project Validation barely uses Cancellation support which leads into trying to Validate outdated document as results into having "strange" error messages logged and exceptions appearance, like #458, #460 and the error messages logged like the  following ones:

```
org.apache.maven.plugin.InvalidPluginDescriptorException: Unable to resolve gins>
			<plugin>
	:exus</groupId>
				<artif
```
or 
```
org.apache.maven.plugin.InvalidPluginDescriptorException: Unable to resolve roupId>org.codehaus:factId>plexus-component-m
```
 
See also: https://github.com/eclipse/lemminx-maven/issues/358

Issue #458, opened 2023-07-27
MavenDiagnosticParticipant returns null range


rgrunber:
```
Jul. 26, 2023 8:29:56 P.M. org.eclipse.lemminx.services.XMLDiagnostics doExtensionsDiagnostics
SEVERE: Error while processing diagnostics for the participant 'org.eclipse.lemminx.extensions.maven.participants.diagnostics.MavenDiagnosticParticipant'.
java.lang.NullPointerException: Property must not be null: range
	at org.eclipse.lsp4j.util.Preconditions.checkNotNull(Preconditions.java:29)
	at org.eclipse.lsp4j.Diagnostic.<init>(Diagnostic.java:91)
	at org.eclipse.lsp4j.Diagnostic.<init>(Diagnostic.java:96)
	at org.eclipse.lsp4j.Diagnostic.<init>(Diagnostic.java:102)
	at org.eclipse.lemminx.extensions.maven.participants.diagnostics.DiagnosticRequest.createDiagnostic(DiagnosticRequest.java:40)
	at org.eclipse.lemminx.extensions.maven.participants.diagnostics.PluginValidator.validateConfigurationElement(PluginValidator.java:154)
	at org.eclipse.lemminx.extensions.maven.participants.diagnostics.PluginValidator.validateConfiguration(PluginValidator.java:95)
	at org.eclipse.lemminx.extensions.maven.participants.diagnostics.MavenDiagnosticParticipant.lambda$doDiagnostics$3(MavenDiagnosticParticipant.java:84)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1850)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #463, opened 2023-07-27
Cancel Project Validation as soon as Diagnostic Request is cancelled …

vrubezhny:
…#462

Fixes: #462

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/PluginValidator.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/PluginValidator.java
@@ -13,6 +13,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -29,24 +30,29 @@ import org.eclipse.lemminx.extensions.maven.utils.MarkdownUtils;
 import org.eclipse.lemminx.extensions.maven.utils.MavenPluginUtils;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 
 class PluginValidator {
 
 	private static final Logger LOGGER = Logger.getLogger(PluginValidator.class.getName());
 
 	private MavenLemminxExtension plugin;
-
-	public PluginValidator(MavenLemminxExtension plugin) {
+	private CancelChecker cancelChecker;
+	
+	public PluginValidator(MavenLemminxExtension plugin, CancelChecker cancelChecker) {
 		this.plugin = plugin;
+		this.cancelChecker = cancelChecker;
 	}
 
-	public Optional<List<Diagnostic>> validatePluginResolution(DiagnosticRequest diagnosticRequest) {
+	public Optional<List<Diagnostic>> validatePluginResolution(DiagnosticRequest diagnosticRequest) throws CancellationException {
+		cancelChecker.checkCanceled();
 		try {
 			MavenPluginUtils.getContainingPluginDescriptor(diagnosticRequest.getNode(), plugin, true);
 		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
 			LOGGER.log(Level.WARNING, "Could not resolve plugin description", e);
 
 			// Add artifactId diagnostic
+			cancelChecker.checkCanceled();
 			String errorMessage = e.getMessage();
 			DOMNode pluginNode = DOMUtils.findClosestParentNode(diagnosticRequest.getNode(), "plugin");
 			Optional<DOMNode> artifactNode = pluginNode.getChildren().stream().filter(node -> !node.isComment())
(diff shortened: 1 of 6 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/VersionValidator.java
```diff
@@ -22,24 +23,35 @@
...
 		Artifact artifact = null;
 		try {
+			cancelChecker.checkCanceled();
 			artifact = new DefaultArtifact(model.getGroupId(), model.getArtifactId(), model.getVersion(), model.getScope(), model.getType(), model.getClassifier(), new DefaultArtifactHandler(model.getType()));
 			if (!artifact.isSelectedVersionKnown()) {
+				cancelChecker.checkCanceled();
 				return Optional.of(Collections
 						.singletonList(diagnosticRequest.createDiagnostic("Version Error", DiagnosticSeverity.Error)));
 			}
 		} catch (Exception e) {
```
angelozerr, line 51: You will catch CancelException here, you need to catch it and rethrow it and not log it.

angelozerr:
The PR will looks very good. Do you see some performance improvement?

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/VersionValidator.java
vrubezhny: It looks like this validator currently isn't used, but you're right.

vrubezhny:
> The PR will looks very good. Do you see some performance improvement?

Not sure if performance improvements are notable (but I'm 100% sure it should be improved), but at least now I don't spot those "strange" exceptions in Diagnostic Participant - that's for sure. 

So I believe this should fix also https://github.com/eclipse/lemminx-maven/issues/458 and https://github.com/eclipse/lemminx-maven/issues/460 issues

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/PluginValidator.java
```diff
@@ -29,24 +30,29 @@
...
 		try {
 			MavenPluginUtils.getContainingPluginDescriptor(diagnosticRequest.getNode(), plugin, true);
 		} catch (PluginResolutionException | PluginDescriptorParsingException | InvalidPluginDescriptorException e) {
 			LOGGER.log(Level.WARNING, "Could not resolve plugin description", e);
 
 			// Add artifactId diagnostic
+			cancelChecker.checkCanceled();
 			String errorMessage = e.getMessage();
 			DOMNode pluginNode = DOMUtils.findClosestParentNode(diagnosticRequest.getNode(), "plugin");
 			Optional<DOMNode> artifactNode = pluginNode.getChildren().stream().filter(node -> !node.isComment())
```
angelozerr, line 58: Do you think it could have some DOMUtils methods which could take time, if yes we should use cancelChecker.

vrubezhny:
@angelozerr  A related question, if you remember,  - does the same https://github.com/eclipse/lemminx/blob/2b0fe29f52c111be3d5c1dccdf9d08340c63be31/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocumentVersionChecker.java#L38 cancel checker works for Code Actions? Or is any other kind of checker used there?

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/PluginValidator.java
vrubezhny: The majority of `DOMUtil` methods are the "short" ones, working with DOM document nodes (like finding children, parents or gathering the text from node value).  So for the most of them the answer is - no, they're pretty quick.
However it worth double-checking the class for possible "long-runners".

angelozerr:
Ok thanks for your feedback. My main question is where lemminx maven is slow? I have the impression is it in parse of MavenProject here https://github.com/vrubezhny/lemminx-maven/blob/c87e0412a1fa82fad6e17f3519d3c2844856809f/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenProjectCache.java#L135 and in this case I wonder if we should manage cancel checjker to stop the builder process ASAP

vrubezhny:
@angelozerr As I answered in slack - it worth to be investigated.

angelozerr:
> @angelozerr As I answered in slack - it worth to be investigated.

Yes sure:) I wanted just to add a comment to remind us about this idea.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/diagnostics/PluginValidator.java
```diff
@@ -29,24 +30,29 @@
...
+	public PluginValidator(MavenLemminxExtension plugin, CancelChecker cancelChecker) {
 		this.plugin = plugin;
+		this.cancelChecker = cancelChecker;
 	}
 
-	public Optional<List<Diagnostic>> validatePluginResolution(DiagnosticRequest diagnosticRequest) {
+	public Optional<List<Diagnostic>> validatePluginResolution(DiagnosticRequest diagnosticRequest) throws CancellationException {
+		cancelChecker.checkCanceled();
 		try {
 			MavenPluginUtils.getContainingPluginDescriptor(diagnosticRequest.getNode(), plugin, true);
```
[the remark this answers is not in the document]
vrubezhny: Done. But I haven't added any new arguments, I've used `node.getOwnerDocument().getCancelChecker()` API

PR #467, opened 2023-07-28
Improve performance with artifact id completion

angelozerr:
This PR improves the completion with artifact id. It seems that completion is slow with artifact ids because the lemminx maven returns too many completion items. In my case I have around 4000 completion items which is slow. 

The idea of this PR is just to filter the artifactid if it matches one of a character of the computed prefix. The second improvement is to compute just one time the prefix range and not for each completion item.

Here a demo before the PR:

LemminxMavenCompletionSlow

Here a demo after the PR

LemminxMavenCompletionPerf


vrubezhny (approved):
The PR looks great!
Thanks!

PR #468, opened 2023-07-28
Initialize maven on background

angelozerr:
This PR initialize maven (plexus container, maven session etc) inside a CompletableFuture which provide the capability to enjoy with syntax validation, xsd validation, completion based on maven xsd without waiting the initialisation of maven.

The second improvement that I did is when we load Maven project from a DOMDocument, I use cancelchecker in the input stream to stop ASAP the build of the MavenProject.


vrubezhny (approved):
The PR looks great!

vrubezhny:
@angelozerr thanks!

PR #470, opened 2023-07-28
Dependency completion performance

angelozerr:
When you open a pom.xml and you trigger completion inside 

```
<dependencies>
   |
```

You should see very quickly `dependency` coming from the XSD, but it doesn't work because:

 * the local  aven searcher is initialized at this step and takes some time
 * or the parse of ModelProject throws a CancellableException which stop the completion process


vrubezhny (approved):
The PR looks good! Thanks!

PR #472, opened 2023-07-29
Use CompletableFuture.completableValue in Junit Test mode

angelozerr:
Use CompletableFuture.completableValue in Junit Test mode

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/LocalRepositorySearcher.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/LocalRepositorySearcher.java
@@ -39,6 +39,8 @@ import org.apache.maven.model.Dependency;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.lemminx.extensions.maven.MavenLemminxExtension;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+import org.eclipse.lsp4j.jsonrpc.CompletableFutures;
 
 public class LocalRepositorySearcher {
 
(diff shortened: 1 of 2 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/LocalRepositorySearcher.java
```diff
@@ -69,78 +71,92 @@ public Collection<Artifact> getLocalPluginArtifacts()  {
...
+					Path localRepoPath = localRepository.toPath();
+					watchService = localRepoPath.getFileSystem().newWatchService();
+					watchKey = localRepoPath.register(watchService, StandardWatchEventKinds.ENTRY_CREATE,
+							StandardWatchEventKinds.ENTRY_DELETE);
+
+					WatchKey key;
+
 					while ((key = (watchService != null ? watchService.take() : null)) != null) {
 						if (watchKey.equals(key)) {
 							cache.remove(localRepository);
```
vrubezhny, line 119: @angelozerr You haven't added any calls to start loading an updated local repo data here. This means a user will most-likely  get an empty artifact list on  next request and local repo loading will be started in background.
Why not to start loading a new local repo right here, so at the moment of new user request it will be propably already loaded and cached?

vrubezhny (commented):
Looks good - there is just a question...

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/LocalRepositorySearcher.java
angelozerr: To be honnest with you, I have just keep the old code which was done in a Thread.

angelozerr: We could improve that in the future.

vrubezhny: OK. Agreed

PR #474, opened 2023-07-29
Stop the thread which collect local repository artifacts

angelozerr:
Stop the thread which collect local repository artifacts


vrubezhny:
@angelozerr Thank you!

PR #475, opened 2023-07-30
Maven project cache cannot be done if one of pom.xml of the workspace doesn't define artifact id.

angelozerr:
Maven project cache cannot be done if one of pom.xml of the workspace
doesn't define artifact id.

If there is one pom.xml in the workspace which doesn't define an artifactid, a version or a groupdid, the call of ArtifactUtils.key and ArtifactUtils.versionLessKey throws an exception and breaks the all code.

For instance it breaks the cache of some pom.xml if ArtifactUtils.key is called with an artifcat which have no group id.


vrubezhny (approved):
Great catch and great fix. Thank you!

Issue #444, opened 2023-07-20
Autocompletion is stuck after some completions


laeubi:
This seem to happen if one use autocompletion "faster" than language-server can process, what I did:

1. I imported https://github.com/eclipse-tycho/tycho/blob/master/ as a project (so there is a toplevel project "tycho")
2. I then open the pom (what is not imported as a project!) https://github.com/eclipse-tycho/tycho/blob/master/demo/osgi-repository/pom.xml
3. I place the cursor under `</configuration>` in a new line
4. I use autocomplete and choose` <executions>`
5. I use autocomplete and choose` <execution>`
6. I use autocomplete and it is completely stuck, two error markers are shown (if I run maven on the commandline it can parse the resulting XML without a problem), closing all editors and open again did not help.

![grafik](https://github.com/eclipse/lemminx-maven/assets/1331477/6865a342-b581-4bb2-ae46-7830c3c307d4)

resulting XML:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<modelVersion>4.0.0</modelVersion>
	<parent>
		<groupId>org.eclipse.tycho.demo</groupId>
		<artifactId>parent</artifactId>
		<version>1.0.0</version>
	</parent>
	<artifactId>repository</artifactId>
	<packaging>repository</packaging>
	<build>
		<plugins>
			<plugin>
				<groupId>org.eclipse.tycho</groupId>
				<artifactId>tycho-repository-plugin</artifactId>
				<version>${tycho-version}</version>
				<!-- Extensions must be enabled for this mojo to work reliable and for using the custom package type-->
				<extensions>true</extensions>
				<configuration>
					<!-- Another option would be using 'local' -->
					<repositoryLayout>maven</repositoryLayout>
				</configuration>
				<executions>
				<execution>
				
				</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
</project>
```

error markers:
```
Description	Resource	Path	Location	Type
Element type "e" must be followed by either attribute specifications, ">" or "/>".	pom.xml	/tycho/demo/osgi-repository/repository	line 28	Language Servers
[...]

PR #483, opened 2023-08-02
Make lazy collectiing of Workspace Artifacts  in completion

vrubezhny:
Make Maven Projects to be loaded/cached lazily when collecting the Workspace Artifacts when gathering completions n order to make content assist faster and more responsive

Issue: #444

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/LoadedMavenProjectProvider.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/LoadedMavenProjectProvider.java
@@ -1,0 +1,124 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lemminx.extensions.maven;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.apache.maven.model.building.FileModelSource;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.extensions.maven.utils.DOMModelSource;
+import org.eclipse.lemminx.extensions.maven.MavenProjectCache.ProjectBuildManager;
+
+import org.eclipse.lemminx.services.IXMLDocumentProvider;
+import org.eclipse.lemminx.utils.FilesUtils;
+
+/**
+ * An object aggregating a build results of a Maven Document, controlling the 
+ * asynchronous access to the Maven Project built from the latest version of 
+ * the provided document.
+ */
+public class LoadedMavenProjectProvider {
+	private static final Logger LOGGER = Logger.getLogger(LoadedMavenProjectProvider.class.getName());
+
+	private final String uri;
+	private final IXMLDocumentProvider documentProvider;
+	private final ProjectBuildManager buildManager;
+
+	private int lastCheckedVersion;
+	private CompletableFuture<LoadedMavenProject> future;
+	
+	/**
+	 * Creates a LoadedMavenProjectProvider using provided URI String identifying the 
+	 * document to be built into a Maven Project. A document found by using Document 
+	 * Provider is the latest version of the document being currently edited or a document 
+	 * read from a file specified by document URI String.
+	 * 
+	 * @param uri A URI String identifying the document
+	 * @param documentProvider An IXMLDocumentProvider instance used to find the latest 
+	 * 		version of the document
+	 * @param buildManager A MavenProject builder
+	 */
+	public LoadedMavenProjectProvider(String uri, IXMLDocumentProvider documentProvider, ProjectBuildManager buildManager) {
+		this.uri = uri;
+		this.documentProvider = documentProvider;
+		this.buildManager = buildManager;
+		this.lastCheckedVersion = -1;
+	}
+	
+	/**
+	 * Returns a `CompletableFuture<LoadedMavenProject>` for asynchronous access
+	 * to the Maven Project built from the latest version of the document.
+	 *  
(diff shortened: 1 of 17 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/LoadedMavenProjectProvider.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.maven.model.building.FileModelSource;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.extensions.maven.utils.DOMModelSource;
+import org.eclipse.lemminx.extensions.maven.MavenProjectCache.ProjectBuildManager;
+
+import org.eclipse.lemminx.services.IXMLDocumentProvider;
+import org.eclipse.lemminx.utils.FilesUtils;
+import org.eclipse.lsp4j.jsonrpc.CompletableFutures;
+
+public class LoadedMavenProjectProvider {
```
angelozerr, line 25: Add Javadoc

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/LoadedMavenProjectProvider.java
```diff
@@ -0,0 +1,104 @@
...
+				source = new FileModelSource(FilesUtils.toFile(uri));
+			}
+			return buildManager.build(uri, source);
+		} catch (Exception e) {
+			LOGGER.log(Level.SEVERE, e.getMessage() + ": " + uri, e);
+			throw e;
+		}
+//		});
+	}
+
```
angelozerr, line 86: Add javadoc to explain that uri can comes from a document (editing) or a file

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -543,6 +558,10 @@ public static boolean match(Path file) {
 				|| fileName.endsWith(Maven.POMv4) || fileName.endsWith(".pom"));
 	}
 
+	public IXMLValidationService getValidationService() {
```
angelozerr, line 561: Add Javadoc

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -765,12 +784,29 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
-	public List<MavenProject> getCurrentWorkspaceProjects() {
+	public List<MavenProject> getCurrentWorkspaceProjects(boolean wait) {
 		return workspaceReader.getCurrentWorkspaceArtifactFiles().stream()
-			.map(f -> getProjectCache().getLastSuccessfulMavenProject(f))
-			.filter(Objects::nonNull).toList();
+					.map(file -> {
+						try {
+							CompletableFuture<LoadedMavenProject> loadedProject = 
+									 getProjectCache().getLoadedMavenProject(toUriASCIIString(file));
+							return wait ? loadedProject.get() : loadedProject.getNow(null);
```
angelozerr, line 801: Please don't never use .get(). us it with a timeout if it is required, but ideally, you shoudl return 

CompletabmeFuture<List<MavenProject>> getCurrentWorkspaceProjects

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
```diff
@@ -765,12 +784,29 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+	 * @param buildIfNecessary A boolean 'true' indicates that all projects are to
+	 *                         be returned, not only the cached ones at the moment,
+	 *                         method should wait for the final build result,
+	 *                         otherwise the only project that are already built and
+	 *                         cached are to be returned, the rest of the projects
+	 *                         are to be built in background
 	 * @return List of Maven Projects
 	 */
-	public List<MavenProject> getCurrentWorkspaceProjects() {
+	public List<MavenProject> getCurrentWorkspaceProjects(boolean wait) {
```
angelozerr, line 795: Please fix javadoc (wait param)

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenLemminxExtension.java
vrubezhny: The argument `wait == true` is used in Rename participant  - when renaming a Maven Property we need to get the full list of child maven projects of the project where the maven property is defined. Currently this support needs a synchronous way of getting such a children list.
In future the rename support can be modified to use  a ` CompletabmeFuture<List>` or other asynchronous functionality. 
I suggest reporting it in a separate issue.

angelozerr: I think it should be safer to use a timeout to avoid blocking thread.

I prefer not having a rename feature than it blocks several threads and the editing support will be slow.

vrubezhny: Added a 10 seconds timeout - do you think it's OK?  (Testing in my local environment is successful with 5 seconds timeout value, but Jenkins CI build is about 1.5 times slower and Windows CI build is about 2 times slower than building in mi local environment, so I'd prefer to leave it that big for now)

mickaelistria:
How is user made aware that the completion results are incomplete because of lazy loading? If there is nothing to warn them, then they may interpret it as a regression.
The fact that some assertions had to change in tests do showcase that this can be perceived as a regression if we don't make it very clear and usable to users that more results are to be provided.

angelozerr:
My scan of my big lcal repo takes arround 15 seconds.

A thing that we could do when completion is triggered while this scan is to provide artifact which are already scanned

angelozerr:
> entry such as "More will be provided after local repo is scanned".

It existed rhis kind of entry but we removed it because completion item is used to apply it and insert some content not to display some i fo

PR #491, opened 2023-08-05
Remote central repository can block Thread pooling

angelozerr:
This PR improves remote central repository : 

 * use 1 cache instead of 3 * 3 caches (just one Thread instead of 9)
 * fix problem with use of synchronized which could bock Thread and don't gree the Thread in the pooling.


vrubezhny (approved):
Looks great! Thanks!

PR #494, opened 2023-08-06
Don't evict the full cache of local repo when m2 repository changes.

angelozerr:
Don't evict the full cache of local repo when m2 repository changes.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
@@ -1,0 +1,145 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lemminx.extensions.maven.searcher.watcher;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.eclipse.lemminx.extensions.maven.searcher.LocalRepositorySearcher;
+
+/**
+ * This class provides the capability to track the deleted/created folder from
+ * the Maven local repository folder to maintain the artifact cache which store
+ * the last version of the artifact.
+ * 
+ * @author Angelo ZERR
+ *
+ */
+public class LocalRepositoryFilesWatcher extends WatchDir {
+
+	private static final Logger LOGGER = Logger.getLogger(LocalRepositoryFilesWatcher.class.getName());
+
+	private final Path localRepoPath;
+	private final Collection<Artifact> cacheArtifacts;
+	private final Thread thread;
+
+	public LocalRepositoryFilesWatcher(Path localRepoPath, Collection<Artifact> cacheArtifacts) throws IOException {
+		super(localRepoPath, true);
+		this.localRepoPath = localRepoPath;
+		this.cacheArtifacts = cacheArtifacts;
+		thread = new Thread(this);
+		thread.setName("Watch Maven local repository '" + localRepoPath + "'");
+		thread.setDaemon(true);
+		thread.start();
+	}
+
+	@Override
+	protected void notifyListeners(WatchEvent<?> event, Path child) {
+		try {
+			boolean checkExistingPomFile = event.kind() == StandardWatchEventKinds.ENTRY_CREATE;
+			DefaultArtifact fileArtifact = LocalRepositorySearcher.createArtifact(child, localRepoPath, null,
+					checkExistingPomFile);
+			if (fileArtifact != null) {
+				// The created / delete folder is an artifact
+				DefaultArtifactVersion fileVersion = new DefaultArtifactVersion(fileArtifact.getVersion());
+				// Get the artifact and version from the cache (which store the last version of
+				// the artifact) for the groupId/artifactId.s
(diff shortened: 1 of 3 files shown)


vrubezhny:
The watcher has the only local repo directory registered, so it may control only the direct children of this directory, not the deeper levels, so it  will not be notified if a new version of an existing  artifact arrives, or a new group is created inside an already existing one.

Currently,  after any change in local repo directory (addind/removing a file or directory) Watcher stops working and exits, probably due  to an exception happening like the following one:
```
Message: Local repo thread watcher event processing error
java.lang.IllegalArgumentException: 'other' is different type of Path
	at java.base/sun.nio.fs.UnixPath.relativize(UnixPath.java:398)
	at java.base/sun.nio.fs.UnixPath.relativize(UnixPath.java:43)
	at org.eclipse.lemminx.extensions.maven.searcher.LocalRepositorySearcher.createArtifact(LocalRepositorySearcher.java:251)
	at org.eclipse.lemminx.extensions.maven.searcher.LocalRepositorySearcher.lambda$getOrCreateLocalArtifactsLastVersion$5(LocalRepositorySearcher.java:143)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:718)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:483)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
```

angelozerr:
@vrubezhny I improved the watcher

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
```diff
@@ -0,0 +1,133 @@
+package org.eclipse.lemminx.extensions.maven.searcher.watcher;
```
vrubezhny, line 1: Add header comment, please

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
```diff
@@ -0,0 +1,133 @@
...
+ * This class provides the capability to track the deleted/created files from
+ * src/main/resources/templates folder of a Qute project and generates the LSP
+ * {@link DidChangeWatchedFilesParams} to update closed document cache.
+ * 
+ * This class is used only when the LSP client cannot support
+ * {@link DidChangeWatchedFilesParams}.
+ * 
+ * @author Angelo ZERR
+ *
+ */
```
vrubezhny, line 28: Please update the JavaDoc comment, as here this class has nothing to do with Qute templates

vrubezhny (commented):
Update the comments please

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
angelozerr: fixed

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/searcher/watcher/LocalRepositoryFilesWatcher.java
angelozerr: fixed

vrubezhny:
I have asked a question on using this Oracle license: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/9848 
I believe we are allowed to incorporate such a class, but if I'n wrong we'll be forced to roll the commit back and replace it with some other solution.

vrubezhny:
When looking for a latest artifact versions a "release" version is to be treated as more recent version comparing to an according "snapshot" or a "qualifier" version, e.g::  `0.9.1 > 0.9.1-SNAPSHOT > 0.9.0`

vrubezhny:
@angelozerr  I've made it to correctly react on different kinds of artifacts deletions (permanent & trash)   and additions  (copy to & building with maven).
Could you, please, give it a try when you have time? 
My changes are added to your branch as a fixup.

vrubezhny:
The PR cannot be applied as it generates errors like:
```
Aug 06, 2023 4:50:52 PM org.eclipse.lemminx.extensions.maven.searcher.LocalRepositorySearcher lambda$getOrCreateLocalArtifactsLastVersion$3
SEVERE: Cannot create local repo thread watcher error
java.io.IOException: User limit of inotify watches reached
	at java.base/sun.nio.fs.LinuxWatchService$Poller.implRegister(LinuxWatchService.java:264)
	at java.base/sun.nio.fs.AbstractPoller.processRequests(AbstractPoller.java:266)
	at java.base/sun.nio.fs.LinuxWatchService$Poller.run(LinuxWatchService.java:364)
	at java.base/java.lang.Thread.run(Thread.java:833)
```
due to exceeding the `fs.inotify.max_user_watches` quota. 
It's not good to force users to modify system parameters, so other solution is to be applied to update the Local Repo Searcher's cache.

mickaelistria:
Isn't the issue caused by the watchService not properly disconnecting the watch(key) before plugging a new one? I think this PR has merits and it could be worth keeping it open and trying to make it work.

vrubezhny:
@mickaelistria Why is this PR restored? It's outdated and doesn't add any value now.

> Isn't the issue caused by the watchService not properly disconnecting the watch(key) before plugging a new one? I think this PR has merits and it could be worth keeping it open and trying to make it work.

We  don't use WatchService anymore for two reasons:
1. To properly update Local Repository Caches it's needed to add a watch key for every directory in the repository (recursively) - it's impossible to do so without modifying the user environment confoguration
2. The old implementation was just dropping the full cache on those changes it have been able to detect - now we're not dropping the cache, but updating the only changed artifacts in case they're just built and cched with lemminx-maven project cache builder and if the local repository is changed from outside (every 30 minutes)

mickaelistria:
I didn't realize that the WatchService was not looking at folder recursively. In another discussion, @angelozerr made me realize that, and I agree with you both that this PR can be closed.
However, some polling can still be useful to update the repo more regularly on changes. I would recommend re-implementing the initially intended, but never working, behavior, using Apache FileAlterationMonitor, as described in https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/monitor/FileAlterationObserver.html

PR #495, opened 2023-08-06
Force the load of the local artifacts done in background.

angelozerr:
Force the load of the local artifacts done in background.


vrubezhny:
@angelozerr Thank you!

PR #502, opened 2023-08-14
Fixed updating the  Local Repository caches

vrubezhny:
Local Repository caches are made to be updated:

- (only temporary lemminx-maven build repository) when a MavenProject is built and cached
- (all cached) every 30 minutes

Also, Diagnostic is made to be re-calculated on finishing of Maven Lemminx Extension initialization

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenInitializationException.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/MavenInitializationException.java
@@ -8,7 +8,7 @@
  *******************************************************************************/
 package org.eclipse.lemminx.extensions.maven;
 
-import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Exception thrown when Maven is initializing.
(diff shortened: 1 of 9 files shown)


mickaelistria:
Wasn't the local repository searcher already placing a file listener to properly update its cache whenever something changes?

angelozerr:
@mickaelistria we removed the thread wich created a watchservice which updated the cache because the original code tracked only the root .m2/repository. I have fixed that to create a watchservice for children folders but we had a problem which created too many watchservice. To avoid this problem we removed this watchservice.

mickaelistria:
OK, got it. And what is used now to detect new content in the Maven repo? I just read about FileAlterationMonitor ( https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/monitor/FileAlterationObserver.html ), would it be capable of being notified recursively?

angelozerr:
For the moment we have no find a solution to evict the cache. 

Just for your information we have 2 local repo now. The user local repo which is not polluated when user is typing in the editor and à .lemminx-maven repo which is polluated

angelozerr:
If it is based on watchservice we will have the same problem, because in short watchservice must never used with a lot of directories

One idea that we had with @vrubezhny is to provide a thread which evicts the cache on each 1 hour

mickaelistria:
FileAlterationMonitor does some polling with a given interval. It can be once per hour or every 3 minutes. Polling a file system isn't too heavy.

mickaelistria:
Here is the code: https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/monitor/FileAlterationMonitor.java

angelozerr:
We indeed try it. Thanks @mickaelistria for your suggestion!

vrubezhny:
I have created https://github.com/eclipse/lemminx-maven/issues/507 to track this.

Issue #505, opened 2023-08-16
Invalid links to line range in external poms in hover


eliasbalasis:
Continuing from #405 :

on Eclipse IDE 2023-03
with M2E 2.4.100-20230815-1219
which contains "M2E - POM Editor using LemMinX language serverlemminx" 2.0.5.20230815-1219
which contains "lemminx-maven" 0.10.1
which contains the fix for #405 

https://github.com/eliasbalasis/eclipse-lemminx-maven-issue-345-root
https://github.com/eliasbalasis/eclipse-lemminx-maven-issue-345-tools
https://github.com/eliasbalasis/eclipse-lemminx-maven-issue-345-project
https://github.com/eliasbalasis/eclipse-lemminx-maven-issue-345-impl

and after importing into Eclipse IDE the above Maven projects in the order mentioned

look at module `eclipse-lemminx-maven-issue-345.tools.retry.springframework.impl.dummy`

Hovering over `<artifactId>eclipse-lemminx-maven-issue-345.tools.retry.springframework.interface</artifactId>` now points to the actual oprigin at `mygroup:eclipse-lemminx-maven-issue-345.tools.retry.springframework:0.0.1-SNAPSHOT` under `<dependencyManagement>` section which is an excellent addition.

However,
hovering over `<artifactId>spring-context</artifactId>` the expected origin now displays which is `org.springframework:spring-framework-bom:5.3.3` as per historical behaviour
![image](https://github.com/eclipse/lemminx-maven/assets/1681838/bb1841f1-cae3-4b2c-976c-ff95d892ec08)
but the link does not jump to the displayed origin.
![image](https://github.com/eclipse/lemminx-maven/assets/1681838/293c24aa-3a19-4622-829f-7c719d830895)

Do you think this is something that could be done and hopefully included in Eclipse 2023-09 release ?

Ideally, the link should be pointing to the actual origin which is `mygroup:eclipse-lemminx-maven-issue-345.tools.retry.springframework:0.0.1-SNAPSHOT`, more precisely the `org.springframework.boot:spring-boot-dependencies` reference under `<dependencyManagement>` section, but this could be a new behaviour, a very useful one in my view though.

vrubezhny:
[...]

PR #511, opened 2023-08-31
Invalid links to line range in external poms in hover #505

vrubezhny:
Fix avoids adding title to links due to problem on Windos OS

Fixes: #505


mickaelistria:
Thank you!

Issue #549, opened 2024-02-13
Doesn't support properties defined in `.mvn/maven.config`


segevfiner:
To create a new structured build/pomless Tycho build for an Eclipse plugin, I have been instructed to look at https://tycho.eclipseprojects.io/doc/4.0.5/StructuredBuild.html, but it seems `.mvn/maven.config` that I'm instructed to create for this is not being picked up by this plugin and as such I get:
```
'build.plugins.plugin.version' for org.eclipse.tycho:target-platform-configuration must be a valid version but is '${tycho-version}'.
```
When opening the pom.xml in Eclipse.

See also https://github.com/eclipse-tycho/tycho/discussions/3481#discussioncomment-8453079

laeubi:
I also noticed this today, this is quite annoying, any hints where properties are evaluated so one can probably suggest a fix?

mickaelistria:
Best is to have a look at Maven code and look for how `maven.config` file is loaded in the session.

laeubi:
> Best is to have a look at Maven code and look for how `maven.config` file is loaded in the session.

Well this happens as part of MavenCLI, m2e already supports it I just don't know how this could / should be applied to lemminx, e.g you can look at

https://github.com/eclipse-m2e/m2e-core/blob/6f195b7171868d6c5d77a16c1aa7c34bdfbb55c7/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenProperties.java#L278-L288

it extracts the data from the maven config given the Multi Module Directory is known like here:

https://github.com/eclipse-m2e/m2e-core/blob/6f195b7171868d6c5d77a16c1aa7c34bdfbb55c7/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenProperties.java#L222-L236

it would be great if this could be fixed because currently users blame this on m2e even though it actually supports this and it is a language server error...

mickaelistria:
Just to not set false expectation: neither Victor nor I are very active on this project at the moment, so it's not likely we try to fix it soon. But as usual, we try to make reviews of incoming PRs highest priority ;)

laeubi:
Wel my problem is that I'm not very familiar with the codebase, but you knwo m2e as well, so maybe you can point me to where something similar like the `MavenExecutionRequest` building happens in lemminx I probably can give it a try :-D

HannesWell:
[...]

PR #552, opened 2024-02-16
Consider properties defined in .mvn/maven.config

laeubi:
Currently lemminx does not consider properties that are defined in the .mvn/maven.config and therefore gives errors or otherwhise wrong results when parsing a project.

This adds support for detecting and reading these properties from the .mvn/maven.config

Fix https://github.com/eclipse/lemminx-maven/issues/549

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
@@ -15,6 +15,10 @@ import java.io.EOFException;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
@@ -23,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
```diff
@@ -364,7 +374,8 @@ public Optional<MavenProject> getSnapshotProject(File file) {
 		 */
 		public MavenProject getSnapshotProject(DOMDocument document, String profileId, boolean resolve) {
 			// it would be nice to directly rebuild from Model instead of re-parsing text
-			ProjectBuildingRequest request = newProjectBuildingRequest(resolve);
+			ProjectBuildingRequest request = newProjectBuildingRequest(resolve, null); // TODO can we get the file from
```
mickaelistria, line 377: new File(document.getDocumentURI())

mickaelistria (changes requested):

mickaelistria:
Thanks, that's a good proposal. Just 1 addition suggested.

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
laeubi: Done

laeubi:
If it is good in general I would polish it a bit (e.g. caching the result of  config parsing maybe), possibly it would make sense to reload the file if save is performed and on typing always use a cached variant I assume that `SnapshotProject` would then always use a cached variant?

mickaelistria:
> If it is good in general I would polish it a bit (e.g. caching the result of config parsing maybe), possibly it would make sense to reload the file if save is performed and on typing always use a cached variant

I think the gain here would be very small because loading projects involves much longer work, so I wouldn't invest in optimizing this particular bit. 

> use a cached variant I assume that SnapshotProject would then always use a cached variant?

I'm not sure I get the question correctly; but my undestanding is that if pom.xml is not the only input (maven.config is too), then the cached project, and it's consumers, needs to be removed from cache when the maven.config changes too. But as it's not a too frequent case to modify maven.config (vs pom.xml), I wouldn't make it a immediate requirement to immediately react on updates.

> Another issue is that a maven.config can provide profiles and different settings, this is currently not taken into account as well but better be solved in a different PR.

OK.

I let you look at what you want to change now, but IMO it's already in a good state to be merged. Please ping whenever you think it's ready to be merged.

laeubi:
@mickaelistria if you are fine with it I'm fine with it as well, I tested it locally and it works as expected for me.

mickaelistria:
Thank you for this very valuable addition!

PR #553, opened 2024-02-16
Consider profiles in the maven.config as well

laeubi:
One can define profiles in the .mvn/maven.config file but currently these are ignored by lemminx leading to unexpected or even fault project models.

This now parses the profile option form the maven config and pass them to the building request.

Like we do in m2e already, @mickaelistria do you can review?

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.StringTokenizer;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -172,8 +173,8 @@ public class MavenProjectCache {
 				DependencyResolutionResult dependencyResolutionResult = null;
 				MavenProject project = null;
 				File file = source.getFile();
-				try {		
-					ProjectBuildingRequest request = newProjectBuildingRequest(file);
+				try {
+					ProjectBuildingRequest request = newProjectBuildingRequest(file, problems);
 
 					ProjectBuildingResult buildResult = projectBuilder.build(source, request);
 					cancelChecker.checkCanceled();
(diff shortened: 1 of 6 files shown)


mickaelistria:
That looks good to me. However, I recommend if possible that you try to add some tests for such features as lemminx-maven is a pretty "mutable" project that's relatively prone to regressions. The good news is that writing tests is very easy.

laeubi:
Can you suggest a simple testcase that checks for a property present in the project? It then should be straight forward to add such test.

mickaelistria:
> that checks for a property present in the project

This has no strong meaning in LSP land, as the grain is Text/UI.
I think you can add a test to Hover, which would hover in some `${property}` which is supposed to show the property value, and create a pom with a profile that sets this value to something like `set-from-profile`, so the output of hover should then show the profile-enabled value for this property by default.

laeubi:
Ok is there any such hover test I can then reuse? Another would be error markers, I can use a version with a property and it should not show any errors...

laeubi:
@mickaelistria I tried to add some test but they failed, as there is not much context beside that it seems I can't get the hover do you have any hint what might be wrong or causing that? Any additional logs I could look at?

mickaelistria:
You can configure Java logging as usual to enable more logging, but can't you just debug those tests? That would be much more efficient.

laeubi:
I now used just the debugger and found that lemminx is not using an `ExpressionEvaluator` for the hover but collecting properties on its own, so I need to enhance this to support user properties as well...

laeubi:
@mickaelistria I got it now working with the test, if this is merged, is there a chance to have a new release of lemminx?
@HannesWell  is currently preparing m2e release and as I understand we need to update the lemminx.jar there to get changes picked up...

lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
```diff
@@ -405,8 +406,8 @@ public MavenProject getSnapshotProject(DOMDocument document, String profileId, b
 		 * 
 		 * @return A ProjectBuildingRequest object
 		 */
-		public ProjectBuildingRequest newProjectBuildingRequest(File projectFile) {
-			return newProjectBuildingRequest(true, projectFile);
+		public ProjectBuildingRequest newProjectBuildingRequest(File projectFile, Collection<ModelProblem> problems) {
```
mickaelistria, line 409: Is the `problems` a "collector" here that will be populated by execution of this method? If so, can you please rename it to `problemsCollector` ?

laeubi: Done.

HannesWell:
@mickaelistria do you plan a new LemMinX release any time soon? I would like to release m2e 2.6.0 actually not later than tomorrow evening and it would be nice if this could be included too.

mickaelistria:
> do you plan a new LemMinX release any time soon?

Done at https://github.com/eclipse/lemminx-maven/releases/tag/0.11.1

HannesWell:
> > do you plan a new LemMinX release any time soon?
> 
> Done at https://github.com/eclipse/lemminx-maven/releases/tag/0.11.1

Thanks for the fast release. I'm about to update m2e to it:
- https://github.com/eclipse-m2e/m2e-core/pull/1692

PR #555, opened 2024-02-19
Use the old deprecated constant for user properties

laeubi:
It seems that under some circumstances using the
CLIManager.SET_USER_PROPERTY badly fails, instead using the deprecated one CLIManager.SET_SYSTEM_PROPERTY fix the problem, so unless the one is really gone use the deprecated field and suppress the warning.

### DIFF
--- a/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
+++ b/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/project/MavenProjectCache.java
@@ -112,6 +112,11 @@ public class MavenProjectCache {
 	}
 	
 	class ProjectBuildManager {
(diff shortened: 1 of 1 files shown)


mickaelistria:
Wouldn't it be simpler to update the Maven embedder in m2e?

laeubi:
> Wouldn't it be simpler to update the Maven embedder in m2e?

it should be the latest version already.... I currently try to find out if it helps, just wanted to make sure nothing breaks for lemminx-maven then...

laeubi:
By the way its a bit unfortunate that in case of fatal error simply nothing seem to haven at all I even don't see any error in the error log :-\

laeubi:
It seems commons.cli is now a requirement for the classpath see:
- https://github.com/eclipse-m2e/m2e-core/pull/1693

mickaelistria:
> It seems commons.cli is now a requirement for the classpath see:

Good catch.

PR #663, opened 2026-06-02
Introduce NoopProgressSupport

manandbytes:
Eliminate null checks for progress' support and monitor.


angelozerr:
Why you want to disable progress support?

manandbytes:
> Why you want to disable progress support?

I don't, this is Null Object pattern so callers don't need null checks before using `progressSupport` or `progressMonitor`.

